PR #38, opened 2021-02-09
feat: Regenerate deviceId function added

yuhao900914:


ios/Classes/SwiftAmplitudeFlutterPlugin.swift
```diff
@@ -93,6 +96,8 @@ import Amplitude
                                                                          withEventProperties: eventProperties,
                                                                          outOfSession: outOfSession)
                     result(true)
+
+                // Event logging
```
haoliu-amp: This comment should be at Ln 90

haoliu-amp (changes requested):

haoliu-amp (approved):

PR #54, opened 2021-06-24
fix: make sure userid is not set to null on startup

keke-dandois:
The iOS native SDK actually provides 2 different init methods, opposed to android. When null is passed on initialization the userid was set to null in the amplitude dashboard, resulting in a wrong flow.

### DIFF
--- a/ios/Classes/SwiftAmplitudeFlutterPlugin.swift
+++ b/ios/Classes/SwiftAmplitudeFlutterPlugin.swift
@@ -34,8 +34,13 @@ import Amplitude
                         return
                     }
 
-                    let userId = args["userId"] as! String?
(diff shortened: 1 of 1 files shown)


yuhao900914:
Hi @keke-dandois 
Thanks for making a contribution to Amplitude-Flutter. 
We will review your pr soon.

yuhao900914 (approved):

spkersten:
@yuhao900914 When do you expect these two PRs be merged? We are more or less waiting for a new version of `amplitude_flutter` that contains these changes.

yuhao900914:
@spkersten We just released the new version. Please check in v3.2.1. 
Thanks.

spkersten:
@yuhao900914 Thank you!

PR #63, opened 2021-09-01
feat: add interface useAppSetIdForDeviceId for android update

qingzhuozhen:


dantetam (approved):
Looks good to me! Checked all the Android versions and API calls, pretty straight forward.

PR #76, opened 2021-10-28
Changed R8 to D8

matyushenko13:
Due to the fact that in gradle = 5.0 r8 deprecated


yuhao900914 (approved):
Thanks @matyushenko13.

PR #79, opened 2021-11-05
Flutter web support structure init + implement amplitude init and logEvent calls

yuhao900914:


lib/amplitude_web.dart
```diff
@@ -0,0 +1,55 @@
...
+          return true;
+        }
+      case "logEvent":
+        {
+          var eventType = args['eventType'];
+          amplitude.logEvent(eventType);
+          return true;
+        }
+      default:
+      //(FlutterMethodNotImplemented)
```
haoliu-amp: throw an exception instead of a comment?

haoliu-amp (changes requested):
The title of the PR makes me feel you finished the bet. :) Could be more descriptive.

example/web/index.html
```diff
@@ -0,0 +1,146 @@
...
+
+  <meta charset="UTF-8">
+  <meta content="IE=Edge" http-equiv="X-UA-Compatible">
+  <meta name="description" content="A new Flutter project.">
+
+  <!-- iOS meta tags & icons -->
+  <meta name="apple-mobile-web-app-capable" content="yes">
+  <meta name="apple-mobile-web-app-status-bar-style" content="black">
+  <meta name="apple-mobile-web-app-title" content="example">
+  <link rel="apple-touch-icon" href="icons/Icon-192.png">
```
qingzhuozhen: Can't find a way to comment on icon files, so just put here, do we need these icons or images?

example/web/index.html
```diff
@@ -0,0 +1,146 @@
...
+
+  <title>example</title>
+  <link rel="manifest" href="manifest.json">
+</head>
+<body>
+
+  <script type="text/javascript" defer>
+    console.log("in lib script snippeit");
```
haoliu-amp: typo.

lib/amplitude_web.dart
```diff
@@ -0,0 +1,59 @@
...
+        {
+          var apiKey = args['apiKey'];
+          amplitude.init(apiKey, null, {});
+          return true;
+        }
+      case "logEvent":
+        {
+          var eventType = args['eventType'];
+          amplitude.logEvent(eventType);
+          return true;
```
qingzhuozhen: For other arguments, is the plan to complete them in PR with other setter methods?

example/web/index.html
haoliu-amp: Do you need this log also

example/web/index.html
```diff
@@ -0,0 +1,146 @@
...
+      (function (e, t) {
+        var n = e.amplitude || { _q: [], _iq: {} }; var r = t.createElement("script")
+          ; r.type = "text/javascript"
+          ; r.integrity = "sha384-9jIJQlEp6cEk6foaEO/N5yzaJmQbTtugbpg9l8P/javk55WlPh7fJGR1vHhyDjON"
+          ; r.crossOrigin = "anonymous"; r.async = true
+          ; r.src = "https://cdn.amplitude.com/libs/amplitude-8.4.0-min.gz.js"
+          ; r.onload = function () {
+            if (!e.amplitude.runQueuedFunctions) {
+        console.log("[Amplitude] Error: could not load SDK")
```
haoliu-amp, line 41: Should it be console.error?

example/web/index.html
```diff
@@ -0,0 +1,146 @@
...
+              loadMainDartJs();
+            }
+          });
+
+        // If service worker doesn't succeed in a reasonable amount of time,
+        // fallback to plaint <script> tag.
+        setTimeout(() => {
+          if (!scriptLoaded) {
+            console.warn(
+              'Failed to load app from service worker. Falling back to plain <script> tag.',
```
haoliu-amp: Prefix with [Amplitude]

example/web/index.html
haoliu-amp: If you do, prefix by [Amplitude]

haoliu-amp (changes requested):

example/web/index.html
yuhao900914: We don't need this actually, That's for my testing. Maybe I will have a code clean-up after all is finished.

lib/amplitude_web.dart
```diff
@@ -0,0 +1,59 @@
...
+    var args = jsonDecode(call.arguments.toString());
+    var instanceName = args['instanceName'];
+    Amplitude amplitude = Amplitude.getInstance(instanceName);
+
+    switch (call.method) {
+      // Init
+      case "init":
+        {
+          var apiKey = args['apiKey'];
+          amplitude.init(apiKey, null, {});
```
qingzhuozhen: this method does have the user id right? Do we want to cover it as well or do it together in implantation PR?

lib/amplitude_web.dart
```diff
@@ -0,0 +1,59 @@
...
+
+import 'package:js/js.dart';
+import 'dart:convert';
+
+import 'package:flutter/services.dart';
+import 'package:flutter_web_plugins/flutter_web_plugins.dart';
+import 'dart:async';
+
+@JS('amplitude')
+class Amplitude {
```
qingzhuozhen: It is ok to keep this here for now. Just wonder if in future we have a long list of methods, should we move this JS interop into another file, then imported into this file?

example/web/index.html
yuhao900914: I copied this from our developer center, do I also need to change it there?

example/web/index.html
```diff
@@ -0,0 +1,146 @@
...
+        // Otherwise, the browser will load the script multiple times,
+        // potentially different versions.
+        var serviceWorkerUrl = 'flutter_service_worker.js?v=' + serviceWorkerVersion;
+        navigator.serviceWorker.register(serviceWorkerUrl)
+          .then((reg) => {
+            function waitForActivation(serviceWorker) {
+              serviceWorker.addEventListener('statechange', () => {
+                if (serviceWorker.state == 'activated') {
+                  console.log('Installed new service worker.');
+                  loadMainDartJs();
```
qingzhuozhen, line 106: Just curious does this work in service worker? I remember we have our customers asked about JS SDK support in service worker.

lib/amplitude_web.dart
yuhao900914: I think we can, will re-organize this.

example/web/index.html
yuhao900914: Actually, that's all auto-generated.  I Will try if we can clean this up.

example/web/index.html
haoliu-amp: If it's from the snippet, that's fine.

lib/amplitude_web.dart
```diff
@@ -0,0 +1,46 @@
...
+  /// https://flutter.dev/go/federated-plugins
+  Future<dynamic> handleMethodCall(MethodCall call) async {
+    var args = jsonDecode(call.arguments.toString());
+    var instanceName = args['instanceName'];
+    Amplitude amplitude = Amplitude.getInstance(instanceName);
+    switch (call.method) {
+      case "init":
+        {
+          String apiKey = args['apiKey'];
+          return amplitude.init(apiKey, args['userId'] ?? null);
```
qingzhuozhen: nit: this default behavior could be moved out of directly calling

lib/amplitude_web.dart
```diff
@@ -0,0 +1,46 @@
...
+        }
+      case "logEvent":
+        {
+          String eventType = args['eventType'];
+          return amplitude.logEvent(eventType, args['eventProperties'] ?? null);
```
qingzhuozhen: Does current example covers event properties as well? If not we can cover it in next PR.

qingzhuozhen (approved):

lib/amplitude_web.dart
yuhao900914: fixed

lib/amplitude_web.dart
yuhao900914: Yes, it's in the example app. The dart Map needs to convert to JSObject before passing into the javascript function. I just made new changes to fix the eventProperties can not be passed into JS.

yuhao900914: Screen Shot 2021-11-09 at 3 54 58 PM

lib/amplitude_web.dart
```diff
@@ -0,0 +1,61 @@
...
+        {
+          String apiKey = args['apiKey'];
+          var userId = args['userId'] ?? null;
+          return amplitude.init(apiKey, userId);
+        }
+      case "logEvent":
+        {
+          String eventType = args['eventType'];
+          var eventProperties = (args['eventProperties'] != null)
+              ? mapToJSObj(args['eventProperties'])
```
qingzhuozhen, line 38: Good found!

qingzhuozhen (approved):

PR #80, opened 2021-11-11
feature: implement functions that don't need more support from Amplitude-JavaScript except Identify/GroupIdentify

yuhao900914:
implement functions that don't need more support from Amplitude-JavaScript except for Identify/GroupIdentify


lib/amplitude_web.dart
```diff
@@ -40,6 +71,81 @@ class AmplitudeFlutterPlugin {
...
+          return amplitude.logRevenue(amount, 1, null);
+        }
+      case "identify":
+        {
+          return true;
+        }
+      case "setGroup":
+        {
+          String groupType = args['groupType'];
+          String groupName = args['groupName'];
```
qingzhuozhen: Is our current flutter interface has group name string only? Noticed in JS surface a dynamic. I believe it is fine if Flutter interface only has String accepted.

qingzhuozhen: Take a quick look, seems that is `dynamic`.

qingzhuozhen (changes requested):

lib/amplitude_web.dart
yuhao900914: Good find. I change the type to dynamic.

haoliu-amp (commented):
The PR title doesn't tell me anything. Could you change it to be more descriptive？

qingzhuozhen (approved):

PR #96, opened 2022-01-17
chore/replacing jcenter

gustavobelo-dev:
Considering jcenter will shut down soon, it is suggested to migrate from jcenter repository to mavenCentral.
Flutter packages are already migrated to it > flutter/flutter#82847

### DIFF
--- a/pubspec.yaml
+++ b/pubspec.yaml
@@ -1,6 +1,6 @@
 name: amplitude_flutter
 description: Amplitude Flutter plugin
-version: 3.8.2
+version: 3.8.1
 homepage: https://www.amplitude.com
 
 environment:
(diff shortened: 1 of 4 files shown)


pubspec.yaml
```diff
@@ -1,6 +1,6 @@
 name: amplitude_flutter
 description: Amplitude Flutter plugin
-version: 3.8.1
```
haoliu-amp: We actually don't need to bump up the version, since it's more for the dev environment.

haoliu-amp (changes requested):
Thanks for your contribution! We only need to update the gradle files.

pubspec.yaml
gustavobelo-dev: Yeah, that makes sense.
Solved > https://github.com/amplitude/Amplitude-Flutter/pull/96/commits/29583a742c02e1f959246de31caaf35b12f4a3e3

haoliu-amp (approved):

Issue #101, opened 2022-04-19
Adjust + Amplitude broken on Android


vishna:
## Expected Behavior
`[adjust] Install` is displayed on Android & iOS

## Current Behavior
`[adjust] Install` is only displayed on  iOS

## Possible Solution
The root of the problem is Android Amplitude SDK, logs show:

> W[/com.amplitude.api.DeviceInfo](https://github.com/together-dev/together/pull/683)( 3416): Google Play Services not available for advertising id
> W[/com.amplitude.api.DeviceInfo](https://github.com/together-dev/together/pull/683)( 3416): Google Play Services SDK not found for app set id!

Seems like `class com.google.android.gms.ads.identifier.AdvertisingIdClient` is not getting found in [DeviceInfo.java](https://github.com/amplitude/Amplitude-Android/blob/2bae5803a076bae8f1e2270909f83fdec3c1912f/src/main/java/com/amplitude/api/DeviceInfo.java#L260)

Followed all the instructions regarding the gradle/proguard setup.

Someone suggested a fix (not using reflection), but afaik it got dismissed https://github.com/dantetam/Amplitude-Android/commit/39c70ab62b54cd67790a05851a5adc81e75d5c7f due to "we would have to change our tests" 🤷‍♀️

Now Adjust SDK seems to find and assign advertising id on Android just fine. My fix introduces option to set custom advertising id, this involves changes in both Android and Flutter repos:

- [Android PR](https://github.com/amplitude/Amplitude-Android/pull/332)
- [Flutter PR](https://github.com/amplitude/Amplitude-Flutter/pull/102)

Then I can simply use it at initialisation time and supply valid advertising id:

```dart
final amplitude = Amplitude.getInstance();
final adId = await Adjust.getGoogleAdId();
await amplitude.setCustomAdvertisingId(advertisingId);
```

This fixed problem we saw in our amplitude dashboard.

## Steps to Reproduce
1. Integrate Flutter Amplitude SDK
2. Integrate Flutter Adjust SDK
3. Publish your apps
4. Observe Adjust install events are reported on iOS while not reported on Android

## Environment
- SDK Version: 3.8.1
- Flutter Version: 2.10 stable
- Device: any Android
- Device OS and Version: any Android

yuhao900914:
Hi @vishna , thanks for contributing our SDKs, we have viewed your pr and left some comments there.

PR #102, opened 2022-04-19
setCustomAdvertisingId

vishna:
solves https://github.com/amplitude/Amplitude-Flutter/issues/101

### DIFF
--- a/lib/amplitude.dart
+++ b/lib/amplitude.dart
@@ -304,6 +304,15 @@ class Amplitude extends _Amplitude {
         'setMinTimeBetweenSessionsMillis', jsonEncode(properties));
   }
 
+  /// Android only, allows to provide a custom advertising id
+  Future<void> setCustomAdvertisingId(String advertisingId) async {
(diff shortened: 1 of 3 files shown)


lib/amplitude.dart
```diff
@@ -304,6 +304,15 @@ class Amplitude extends _Amplitude {
         'setMinTimeBetweenSessionsMillis', jsonEncode(properties));
   }
 
+  /// Android only, allows to provide a custom advertising id
+  Future<void> setCustomAdvertisingId(String advertisingId) async {
```
qingzhuozhen, line 308: Wonder if need some placeholder method in iOS and web @yuhao900914

yuhao900914: Yes, that's right. Both ios and web need a placeholder otherwise it's not able to build.  @vishna  Can you please update this?

yuhao900914 (changes requested):
Need to add a function mapping placeholder in both ios and android.

PR #108, opened 2022-05-02
Run `flutter pub get` in example

nilsreichardt:
I used Flutter version `2.10.5`

### DIFF
--- a/example/pubspec.lock
+++ b/example/pubspec.lock
@@ -14,7 +14,7 @@ packages:
       name: async
       url: "https://pub.dartlang.org"
     source: hosted
-    version: "2.8.1"
+    version: "2.8.2"
   boolean_selector:
     dependency: transitive
     description:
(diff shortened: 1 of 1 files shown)


example/pubspec.lock
```diff
@@ -92,7 +92,14 @@ packages:
       name: matcher
       url: "https://pub.dartlang.org"
     source: hosted
-    version: "0.12.10"
+    version: "0.12.11"
+  material_color_utilities:
```
yuhao900914, line 96: Is this dependency necessary?

yuhao900914, line 96: Is this dependency necessary?

nilsreichardt: It's a dependency by the Flutter SDK which have been added some months ago: https://github.com/flutter/flutter/pull/94377. I have not added any dependency. I just ran `flutter pub get` with the newest stable version in the example directory.

yuhao900914: Got it. Thanks.

PR #110, opened 2022-05-11
chore: update amplitude android version

yuhao900914:
The android app set id has been updated in amplitude-android. Adopt the newest amplitude-android in order to reflect the change in amplitude-flutter.

### DIFF
--- a/example/android/build.gradle
+++ b/example/android/build.gradle
@@ -1,5 +1,5 @@
 buildscript {
-    ext.kotlin_version = '1.3.40'
+    ext.kotlin_version = '1.6.0'
     repositories {
         google()
         mavenCentral()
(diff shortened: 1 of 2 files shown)


example/android/build.gradle
```diff
@@ -1,5 +1,5 @@
 buildscript {
-    ext.kotlin_version = '1.3.40'
+    ext.kotlin_version = '1.6.0'
```
qingzhuozhen, line 2: Is this for the upgrade?

yuhao900914: Have the same issue when building the android app like this https://github.com/flutter/flutter/issues/83834 
I think the issue is caused by I updated my android studio recently and the current kotlin plugin version is 1.6.0 which is not compatible with kotlin version (1.3.40) in the project. 
It shouldn't be an issue for other customers.

qingzhuozhen (approved):

PR #123, opened 2022-10-21
chore: update the kotlin gradle version for compatible with latest gradle version

### LINKED ISSUE #119: Not compatible with latest android gradle plugin
The existing gradle dependency has kotlin version `1.3.40` which is not compatible with latest gradle and kotlin plugin. Please upgrade it to latest version of kotlin plugin

yuhao900914:
Description:

The existing gradle dependency has kotlin plugin 1.3.40 which is not compatible with the latest gradle and kotlin plugin. 
After discussion, we decided to support the latest version of gradle and recommend the customer who use lower version of the gradle to use lower version of amplitude flutter.

Upgrade versions based on https://kotlinlang.org/docs/whatsnew17.html#bumping-minimum-supported-versions

This is the related PR for the doc.
https://github.com/Amplitude-Developer-Docs/amplitude-dev-center/pull/345

### DIFF
--- a/android/build.gradle
+++ b/android/build.gradle
@@ -2,14 +2,14 @@ group 'com.amplitude.amplitude_flutter'
 version '1.0-SNAPSHOT'
 
 buildscript {
-    ext.kotlin_version = '1.3.40'
+    ext.kotlin_version = '1.7.10'
     repositories {
         google()
         mavenCentral()
     }
 
     dependencies {
-        classpath 'com.android.tools.build:gradle:3.2.1'
+        classpath 'com.android.tools.build:gradle:3.6.4'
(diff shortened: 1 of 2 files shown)


liuyang1520 (approved):

example/android/build.gradle
```diff
@@ -1,5 +1,5 @@
 buildscript {
-    ext.kotlin_version = '1.6.0'
+    ext.kotlin_version = '1.7.10'
```
qingzhuozhen: This is just the example, right? Won't really affect anything in our SDK

yuhao900914: Yes.

qingzhuozhen: OK. I guess we are not bumping up version after this change? More like to have another PR?

qingzhuozhen (approved):

PR #139, opened 2023-04-04
feat: add setEventUploadPeriodMillis support

justin-fiedler:
Summary
* Added new `setEventUploadPeriodMillis` method to configure underlying SDKs
* Should support Android, iOS, and web
* Web implementation is questionable, better implementation would require updates to `amplitude-js`
* Added "Flush Intervals" section to the example app to test "setEventUploadThreshold" and "setEventUploadPeriodMillis"


qingzhuozhen (approved):

yuhao900914 (commented):
For the web, we also need to add the function signature to this file. 
lib/web/amplitude_js.dart

yuhao900914 (approved):

PR #142, opened 2023-04-18
Expose "setOffline" method

### LINKED ISSUE #81: Expose setOffline
## Summary
Expose `setOffline` via the flutter plugin

Both objective-c and Kotlin counterparts offer a method 'setOffline'.

```
/**
     * Sets offline. If offline is true, then the SDK will not upload events to Amplitude servers;
     * however, it will still log events.
     *
     * @param offline whether or not the SDK should be offline
     * @return the AmplitudeClient
     */
    public AmplitudeClient setOffline(boolean offline) {
        this.offline = offline;

        // Try to update to the server once offline mode is disabled.
        if (!offline) {
            uploadEvents();
        }

        return this;
    } 
``` 

```
- (void)setOffline:(BOOL)offline {
    _offline = offline;

    if (!_offline) {
        [self uploadEvents];
    }
}
```

## Motivations

Prevent uploading events to Amplitude server in special conditions.

nishiths23:
Expose the "setOffline" method, already available on the Android/iOS Amplitude SDKs.

Closes the following feature request: https://github.com/amplitude/Amplitude-Flutter/issues/81

### DIFF
--- a/lib/amplitude_web.dart
+++ b/lib/amplitude_web.dart
@@ -179,7 +179,9 @@ class AmplitudeFlutterPlugin {
           return amplitude.setServerZone(serverZone, updateServerUrl);
         }
       case "setOffline":
-        return false;
+        {
+          return false;
+        }
       default:
         throw PlatformException(
           code: 'Unimplemented',
--- a/lib/amplitude_web.dart
+++ b/lib/amplitude_web.dart
@@ -1,10 +1,11 @@
+import 'dart:async';
 import 'dart:convert';
 
 import 'package:flutter/services.dart';
 import 'package:flutter_web_plugins/flutter_web_plugins.dart';
 import 'package:js/js_util.dart' as js;
+
 import 'web/amplitude_js.dart';
-import 'dart:async';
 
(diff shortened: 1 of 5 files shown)


nishiths23:
@yuhao900914 tagging maintainers for review

lib/amplitude_web.dart
```diff
@@ -177,6 +178,8 @@ class AmplitudeFlutterPlugin {
           bool updateServerUrl = args['updateServerUrl'];
           return amplitude.setServerZone(serverZone, updateServerUrl);
         }
+      case "setOffline":
+        return false;
```
justin-fiedler: Please use standard code style
```
case "setOffline":
  {
    return false;
  }
```

justin-fiedler (commented):
Thanks @nishiths23 for the feature. Overall looks good, see nit about code style.

yuhao900914 (approved):

lib/amplitude_web.dart
nishiths23: Updated

PR #151, opened 2023-07-21
Fix #150 - include nested objects in mapToJSObj

### LINKED ISSUE #150: Nested event_properties maps only serialize the top level on web
## Expected Behavior
When calling `logEvent` on web with a nested `eventProperties` argument, I expect that whole object to be converted to a json string, not just the top level properties 

## Current Behavior
When calling `logEvent` with a nested `eventProperties` argument, any object properties at the top level get converted to empty json objects. For example, if I sent the following map: `{"topLevel1":"string","topLevel2":0,"topLevel3": {"nested":true}}`, the event ends up getting logged as `{"topLevel1":"string","topLevel2":0,"topLevel3": {}}` (note `topLevel3` is reduced to an empty object)

## Possible Solution
I believe this can be fixed by changing `mapToJSObj` in `amplitude_web.dart` to be recursive. For example, changing line 193 from `var value = v;` to `var value = (v is Map) ? mapToJSObj(v) : v;`

## Steps to Reproduce
1. Create a new flutter project for web
2. set up amplitude sdk
3. call `logEvent` with an `eventProperties` argument that includes nested maps/objects
4. observe network traffic to see that all objects are reduced to an empty object in the actual payload of the request to amplitude

## Environment
- SDK Version: 3.15.0
- Flutter Version: 
- Device: Web (reproduced on a macbook)
- Device OS and Version: macOS 13.3.1, running Chrome 114

jtothebell:
This should fix issue #150 where nested objects are not included when sending events with nested `eventProperties` objects by recursively calling `mapToJSObj` if the value is a map.

### DIFF
--- a/lib/amplitude_web.dart
+++ b/lib/amplitude_web.dart
@@ -190,7 +190,7 @@ class AmplitudeFlutterPlugin {
     var object = js.newObject();
     map.forEach((k, v) {
       var key = k;
-      var value = v;
+      var value = (v is Map) ? mapToJSObj(v) : v;
       js.setProperty(object, key, value);
     });
     return object;


justin-fiedler (approved):

justin-fiedler:
Thank you @jtothebell!

PR #156, opened 2023-10-08
feat: add support for macos

### LINKED ISSUE #46: Flutter macOS Support
## Description
Hi! Our Flutter app [Sharezone](https://sharezone.net) is available as an Android, iOS, Web and macOS app. We would love to see macOS support ❤️

jointhejourney:
Hi there,

Our Flutter app, [Journey Navigation](https://gowithjourney.com), utilizes Amplitude for our iOS app and we wanted to use it for our MacOS app as well.

I took advantage of the `sharedDarwinSource` option, since the Amplitude-iOS package is compatible with MacOS. The changes are fairly straightforward, and everything seems to work ok.

Fixes #46


jointhejourney:
One thing I did notice, is that "Device Type" isn't available when running on MacOS. However, I don't think it has to do with this Flutter package, it might be an issue with Amplitude-iOS.

edit:

I tried overriding the event from the Flutter side of things by using the `device_info_plus` package + `apple_product_name` but couldn't get it to work. Seems like it can only be fixed by the Amplitude-iOS package.

jointhejourney:
@justin-fiedler any chance we could get this merged in?

izaaz:
@jointhejourney Thanks for created this pull request. We are in the process of updating the flutter SDK to use our newer iOS and Android SDKs. So ideally, we'd want to use the macos support in our swift SDK. I'll keep this thread updated on an ETA on when this will be made available.

lukemmtt:
@chungdaniel I believe this can be closed now, since #46 is closed

chungdaniel:
Closing as #217 adds macOS support to 4.x beta

PR #166, opened 2024-02-07
chore: add identify and base event

Mercy811:
## Description
- Add `Identify` and `BaseEvent` 
- The old `/lib/identify.dart` is still there for dev reference, I will remove it after updating the main Amplitude interfaces like `amplitude.identify()`.

## User interfaces
```
final identify = Identify()
    ..set(property: "color", value: "green")
    ..setOnce(property: "property", value: "value")
    ..add(property: "property", value: "value");

// Event type is required
final event1 = BaseEvent(eventType: "button clicked");
// Can also pass other fields when creating a base event
final event2 = BaseEvent(eventType: "button clicked", eventProperties: {"button color": "green"});
```

## Tests
image

### identify
Some logs in identify are not covered. Didn't find an easy way to intercept logging stream or mock and replace the logger. 
image


lib/constants.dart
```diff
@@ -1,6 +1,7 @@
 class Constants {
   static const packageName = 'amplitude-flutter';
   static const packageVersion = '3.16.1';
+  static const identify_event = "\$identify";
```
justin-fiedler, line 4: nit: I think `identify_event_type` is more explicit, and could help prevent customer confusion if they try to send a "identify event"

lib/events/base_event.dart
```diff
@@ -0,0 +1,188 @@
...
+  String eventType;
+  Map<String, dynamic>? eventProperties;
+  Map<String, dynamic>? userProperties;
+  Map<String, dynamic>? groups;
+  Map<String, dynamic>? groupProperties;
+
+  BaseEvent({
+    String? userId,
+    String? deviceId,
+    int? timestamp,
```
justin-fiedler, line 15: nit: I think it would be good to put the more common properties up top - e.g. eventType, eventProperties and put the ones that we normally set automatically near the bottom. Or alphabetically.

lib/events/base_event.dart
```diff
@@ -0,0 +1,188 @@
...
+    Plan? plan,
+    IngestionMetadata? ingestionMetadata,
+    double? revenue,
+    double? price,
+    int? quantity,
+    String? productId,
+    String? revenueType,
+    Map<String, dynamic>? extra,
+    String? partnerId,
+    int attempts = 0,
```
justin-fiedler: Not sure we should expose this internal value to customers (retry attempts)

lib/events/base_event.dart
```diff
@@ -0,0 +1,188 @@
...
+      'product_id': productId,
+      'revenue_type': revenueType,
+      'extra': extra,
+      'partner_id': partnerId,
+      'attempts': attempts,
+    };
+  }
+
+  void mergeEventOptions(EventOptions options) {
+    if (options.userId != null) userId = options.userId!;
```
justin-fiedler: nit: null coalescing could make this a little easier to read e.g. `userId = options.userId ?? userId` https://stackoverflow.com/a/17006968/5046784

lib/events/identify.dart
```diff
@@ -0,0 +1,103 @@
...
+  static const IdentifyOperation unset = IdentifyOperation._("\$unset");
+  static const IdentifyOperation preInsert = IdentifyOperation._("\$preInsert");
+  static const IdentifyOperation postInsert = IdentifyOperation._("\$postInsert");
+  static const IdentifyOperation remove = IdentifyOperation._("\$remove");
+}
+
+class Identify {
+  Set<String> propertySet = {};
+  Map<String, dynamic> properties = {};
+  final log = Logger('Amplitude-Identify');
```
justin-fiedler, line 23: Not super familar with Dart, but should we be using the logger passed in by the Configuration instead?

lib/constants.dart
Mercy811: I have the same feeling but this is what is used in [Kotlin](https://github.com/amplitude/Amplitude-Kotlin/blob/6402b2a8958bb6fc9e739c771da171295ad3f144/core/src/main/java/com/amplitude/core/Constants.kt#L10) and [Swift](https://github.com/amplitude/Amplitude-Swift/blob/6e8c1b44b30e15237d5c3bc172e7227672bd56b1/Sources/Amplitude/Constants.swift#L54). I am slightly leaning towards keep aligned.

lib/events/base_event.dart
Mercy811: These are named parameters so I don't feel manually order them matters. Also it's already ordered automatically by the [Dart plugin](https://docs.flutter.dev/tools/formatting#:~:text=Automatically%20formatting%20code%20in%20Android%20Studio%20and%20IntelliJ,-Install%20the%20Dart&text=To%20format%20your%20code%20in,press%20Ctrl%20%2B%20Alt%20%2B%20L%20.).

lib/events/identify.dart
Mercy811: Identify class doesn't have a amplitude or configuration. See [Kotlin](https://github.com/amplitude/Amplitude-Kotlin/blob/60120f281a02c9abf7c90176135aff9f95c0e801/core/src/main/java/com/amplitude/core/events/Identify.kt#L562) and [Swift](https://github.com/amplitude/Amplitude-Swift/blob/88c6e30769453697d5080b6ea7579e0a0ec4e1e9/Sources/Amplitude/Events/Identify.swift#L30)

lib/events/base_event.dart
Mercy811: Removed it from the constructor at https://github.com/amplitude/Amplitude-Flutter/pull/166/commits/46d8447c787ed67226769c73e89855c08af0d333

lib/events/base_event.dart
Mercy811: Good call out! Dart introduced null coalescing in 2.12 which is the minimum version our SDK supports. Updated at https://github.com/amplitude/Amplitude-Flutter/pull/166/commits/46d8447c787ed67226769c73e89855c08af0d333

justin-fiedler (approved):

yuhao900914:
@Mercy811 
Maybe we can skip the log?
https://stackoverflow.com/a/23264375

Mercy811:
Thanks @yuhao900914 I will add it in the next PR.

Mercy811:
It's for C++ and doesn't work in Dart 😔

yuhao900914:
Good to know. Thanks Xinyi.

PR #162, opened 2023-12-18
refactor: add configuration

Mercy811:
Add Configuration class. Configuration is part of Amplitude class and will be used later in Amplitude class. 

`toMap()` is a serializer to stringify Configuration object before passing it from Flutter to native platforms via a method channel.

### Configuration
```
var configuration = Configuration(apiKey: "apiKey") // api key is required

var configuration = Configuration(
  apiKey: "apiKey",
  enableCoppaControl: true, // mobile
  useAdvertisingIdForDeviceId: true, // Android
  appVersion: '1.0.0', // web
  trackingOptions: TrackingOptions(
    ipAddress: true,
    country: true, // mobile
    adid: true, // Android
    idfv: true, // iOS
    ),
  defaultTracking: DefaultTrackingOptions(
    sessions: false,
    appLifecycles: false, // mobile
    deepLinks: false, // Android
    attribution: false, // web
  ),
);

// Default tracking options shortcuts
// Disable all
var configuration = Configuration(
    apiKey: "apiKey",
    defaultTracking: DefaultTrackingOptions.none()
);

// Enable all
var configuration = Configuration(
    apiKey: "apiKey",
    defaultTracking: DefaultTrackingOptions.all()
);

```

## Tests
Run the following command to generate test coverage file 
```
./generate_test_coverage.sh 
```
image

The code is based on the comparison of configurations supported in latest iOS, Android and Browser SDKs
https://docs.google.com/spreadsheets/d/1ZFrXjzdfMWw04a8JSUsPrljGM5CTVQSY0XBBiEK5DfY/edit#gid=0

Migration guide:
https://amplitude.atlassian.net/wiki/spaces/GOV/pages/2262892810/Migration+guide#Configuration

PR #168, opened 2024-02-12
refactor: update Amplitude interfaces

Mercy811:
Comment out `example/`. Will update it in a separate ticket https://amplitude.atlassian.net/browse/AMP-89883

## Update Amplitude interfaces. 

See [migration guide](https://amplitude.atlassian.net/wiki/spaces/GOV/pages/2262892810/Migration+guide) for details.

Some interfaces are not supported as Dart's doesn't support overloading. But workarounds/similar are provided. 
- `track({required String eventType, dynamic eventProperties, EventOptions? options})` is not supported, but can be replaced by `track({required BaseEvent event, EventOptions? options,})`

See a full list of interfaces comparison with Kotlin and Swift SDK [here](https://docs.google.com/spreadsheets/d/1EsjlwObtqObWKeXUHh81xXxf8VrPD0QxA_I3q3roYF4/edit#gid=0).

## Refactor TrackingOptions
- Refactor toMap() to remove `disableFields` to reflect the style of tracking option config (replace `disable` and `enable` prefix with boolean values).
- Remove corresponding constants.
image


example/lib/my_app.dart
```diff
@@ -32,35 +32,35 @@ class _MyAppState extends State<MyApp> {
...
-      'buildNumber': '1.1.1',
-    };
-    analytics.logRevenueAmount(21.9);
-    analytics.setUserProperties(userProps);
+    // Map<String, dynamic> userProps = {
+    //   'date': '01.06.2020',
+    //   'name': 'Name',
+    //   'buildNumber': '1.1.1',
+    // };
+    // analytics.logRevenueAmount(21.9);
```
justin-fiedler, line 58: Instead of commenting these out it might be good update the app as you go to test new interface

justin-fiedler (approved):
Thanks @Mercy811!

example/lib/my_app.dart
Mercy811: > Comment out example/. Will update it in a separate ticket https://amplitude.atlassian.net/browse/AMP-89883

Yes, like the PR description says, I have a separate ticket for it. I feel unit tests are enough for now.

example/lib/deviceid_sessionid.dart
```diff
@@ -16,29 +16,29 @@ class _DeviceIdSessionIdState extends State<DeviceIdSessionId> {
...
-          children: [
-            FutureBuilder(
-              future: AppState.of(context).analytics.getDeviceId(),
-              builder: (BuildContext context, AsyncSnapshot<dynamic> snapshot) {
-                return Text(snapshot.data.toString());
-              },
-            ),
-          ],
-        ),
+        // Row(
```
yuhao900914, line 19: Why not comment on multiple lines together?

lib/amplitude.dart
```diff
@@ -1,342 +1,211 @@
...
 
-  /// Defines a custom session expiration time where the timeout input is in milliseconds.
-  Future<void> setMinTimeBetweenSessionsMillis(int timeInMillis) async {
-    Map<String, dynamic> properties = _baseProperties();
-    properties['timeInMillis'] = timeInMillis;
+  /// Sets a custom device ID.
+  ///
+  /// Make sure the value is sufficiently unique. Amplitude recommends using a UUID.
+  Future<void> setDeviceId(String deviceId) async {
+    Map<String, dynamic> properties = {};
```
yuhao900914, line 194: Is this needed?

lib/tracking_options.dart
```diff
@@ -67,38 +65,27 @@ class TrackingOptions {
...
-    if (!deviceBrand) disableFields.add(Constants.ampTrackingOptionDeviceBrand);
-    if (!latLag) disableFields.add(Constants.ampTrackingOptionLatLag);
-    if (!apiLevel) disableFields.add(Constants.ampTrackingOptionApiLevel);
-
-    // iOS-specific tracking options
-    if (!idfv) disableFields.add(Constants.ampTrackingOptionIdfv);
-
+  Map<String, bool> toMap() {
     return {
-      'disabledFields': disableFields,
```
yuhao900914, line 101: How we disable the tracking options now?

example/lib/deviceid_sessionid.dart
Mercy811: I commented the future line first and then realized that I need to comment out the whole Row. I don't think this matters a lot because I will update the example in a separate ticket later anyway.

lib/tracking_options.dart
Mercy811: It follows browser SDK pattern. See the tests at 'test/tracking_options_test.dart' and more context around the change [here](https://github.com/amplitude/Amplitude-Flutter/pull/162#discussion_r1478923413).

lib/amplitude.dart
Mercy811: I'd like to follow the pattern and keep the argument in `invokeMethod<T>(String method, [dynamic arguments])` to be a json string so that all messages can be decoded on clients regardless of `method`.

lib/tracking_options.dart
yuhao900914: Thanks for the content.

yuhao900914 (approved):

PR #171, opened 2024-02-23
refactor: update Android bridge

Mercy811:
- Update Android bridge file at `android/src/main/kotlin/com/amplitude/amplitude_flutter/AmplitudeFlutterPlugin.kt`
- Add Android native side tests at `android/src/test/kotlin/com/amplitude/amplitude_flutter/AmplitudeFlutterPluginTest.kt`
  - Simulate method call from Flutter 
  - Verify corresponding Android APIs are called
 

To run Android native unit tests `cd example/android && ./gradlew testDebugUnitTest`
Unfortunately I cannot make Android unit test CI running. Will just keep the tests to local for now. 
image


android/build.gradle
```diff
@@ -2,14 +2,14 @@ group 'com.amplitude.amplitude_flutter'
 version '1.0-SNAPSHOT'
 
 buildscript {
-    ext.kotlin_version = '1.7.10'
+    ext.kotlin_version = '1.5.20'
```
izaaz, line 5: Are we going to an earlier version of kotlin?

lib/amplitude.dart
```diff
@@ -147,7 +147,7 @@ class Amplitude extends _Amplitude {
       event.mergeEventOptions(options);
     }
 
-    return await _channel.invokeMethod("track", jsonEncode(event.toMap()));
```
izaaz, line 150: Why does this change? Was the old version calling the wrong method or did the interface change in the new Android SDK?

izaaz (approved):

android/build.gradle
```diff
@@ -44,6 +44,20 @@ android {
 }
 dependencies {
     implementation "org.jetbrains.kotlin:kotlin-stdlib-jdk7:$kotlin_version"
-    implementation "com.amplitude:android-sdk:2.39.8"
-    implementation "com.squareup.okhttp3:okhttp:4.2.2"
```
yuhao900914, line 48: Are we going to implement the migration logic? Or should that be done in other tickets?

android/src/main/kotlin/com/amplitude/amplitude_flutter/AmplitudeFlutterPlugin.kt
```diff
@@ -263,236 +139,157 @@ class AmplitudeFlutterPlugin : FlutterPlugin, MethodCallHandler {
...
 
-                    // UNSET
-                    "\$unset" -> {
-                        identify.unset(key)
-                    }
+    internal fun getEvent(json: JSONObject): BaseEvent {
+        val plan = json.getJSONObject("plan")
+        val ingestionMetadata = json.getJSONObject("ingestion_metadata")
+        val event = BaseEvent()
+        event.eventType = json.getString("event_type")
```
yuhao900914, line 203: Do we need to check if  the input has event_type or not?

android/build.gradle
Mercy811: What is the migration logic for?

lib/amplitude.dart
Mercy811: This is to fix the error of my last commit. Note this PR is compared with branch `4.x` not `main`. The interface `setGroup()` remains unchanged.

android/src/main/kotlin/com/amplitude/amplitude_flutter/AmplitudeFlutterPlugin.kt
Mercy811: `eventType` is a required for `BaseEvent` constructor. So skip checking whether it's existing or not.

android/build.gradle
Mercy811: I think I lowered it for some compatibility issues. I don't see any harms to customers to lower the version. Do you any concern? I'm open to discussion and can try to see if I can revert it if it's necessary.

android/build.gradle
yuhao900914: The migration from the older flutter to the latest flutter.

Mercy811: Data migration in Kotlin SDK is turned on by default https://www.docs.developers.amplitude.com/data/sdks/android-kotlin/migration/#data-migration. Could you elaborate on the what we need to migrate?

yuhao900914: NVM. The flutter is a bridge of the latest ios sdk and latest android sdk. The migration is over there.

yuhao900914 (approved):

android/build.gradle
```diff
@@ -25,7 +25,7 @@ apply plugin: 'com.android.library'
 apply plugin: 'kotlin-android'
 
 android {
-    compileSdkVersion 28
+    compileSdkVersion 31
```
justin-fiedler: Since we are updating any reason not to use latest? e.g. 34

android/build.gradle
```diff
@@ -44,6 +44,20 @@ android {
 }
 dependencies {
     implementation "org.jetbrains.kotlin:kotlin-stdlib-jdk7:$kotlin_version"
-    implementation "com.amplitude:android-sdk:2.39.8"
-    implementation "com.squareup.okhttp3:okhttp:4.2.2"
+    implementation "org.jetbrains.kotlin:kotlin-reflect:1.6.10"
+
+    implementation 'com.amplitude:analytics-android:1.+'
```
justin-fiedler: nit: This sometimes gives warnings I have switched to Maven style ranges instead

https://github.com/amplitude/session-replay-android/blob/main/plugin-session-replay/build.gradle.kts#L42

android/build.gradle
```diff
@@ -2,14 +2,14 @@ group 'com.amplitude.amplitude_flutter'
...
-    ext.kotlin_version = '1.7.10'
+    ext.kotlin_version = '1.5.20'
     repositories {
         google()
         mavenCentral()
     }
 
     dependencies {
-        classpath 'com.android.tools.build:gradle:3.6.4'
+        classpath 'com.android.tools.build:gradle:7.0.4'
```
justin-fiedler: If we are updating can we use latest? e.g. gradle 8.2.x

android/src/main/kotlin/com/amplitude/amplitude_flutter/AmplitudeFlutterPlugin.kt
```diff
@@ -1,260 +1,136 @@
...
+                val defaultTracking = json.getJSONObject("defaultTracking")
+                amplitude = Amplitude(
+                    Configuration(
+                        apiKey = json.getString("apiKey"),
+                        context = ctxt!!,
+                        flushQueueSize = json.getInt("flushQueueSize"),
+                        flushIntervalMillis = json.getInt("flushIntervalMillis"),
+                        instanceName = json.getString("instanceName"),
+                        optOut = json.getBoolean("optOut"),
+                        minIdLength = if (json.optInt("minIdLength") == 0) defaultMinIdLength else json.optInt(
```
justin-fiedler:
```suggestion
                        minIdLength = if (json.optInt("minIdLength") < 1) defaultMinIdLength else json.optInt(
```

android/src/main/kotlin/com/amplitude/amplitude_flutter/AmplitudeFlutterPlugin.kt
```diff
@@ -263,236 +139,157 @@ class AmplitudeFlutterPlugin : FlutterPlugin, MethodCallHandler {
...
-                    }
-                }
+    internal fun JSONObject.toMutableMap(): MutableMap<String, Any?> {
+        val map = mutableMapOf<String, Any?>()
+        val keys = keys()
+        while (keys.hasNext()) {
+            val key = keys.next()
+            var value = get(key)
+            if (value is JSONObject) {
+                value = value.toMap()
```
justin-fiedler, line 275: Should this be mutable too? e.g. `value.toMutableMap()`

example/android/build.gradle
```diff
@@ -1,12 +1,12 @@
...
-    ext.kotlin_version = '1.6.0'
+    ext.kotlin_version = '1.6.21'
     repositories {
         google()
         mavenCentral()
     }
 
     dependencies {
-        classpath 'com.android.tools.build:gradle:4.0.1'
+        classpath 'com.android.tools.build:gradle:7.4.2'
```
justin-fiedler: Can we update this to match the gradle version used by the SDK?

justin-fiedler (approved):

android/build.gradle
Mercy811: Updated to 34 . Where do you normal check the latest version? I searched "compileSdkVersion latest version" but didn't get an answer to be 34.

android/build.gradle
Mercy811: `7.0.4` was the suggested version by Android Studio somehow. Updated to `classpath 'com.android.tools.build:gradle:8.2.2'`.

android/build.gradle
Mercy811: Update to `implementation 'com.amplitude:analytics-android:[1.0,2.0)'`

android/src/main/kotlin/com/amplitude/amplitude_flutter/AmplitudeFlutterPlugin.kt
Mercy811: This function will be removed in my next PR.

example/android/build.gradle
Mercy811: Updated to 8.2.2.

android/build.gradle
justin-fiedler: Thanks. I knew 34 was latest from the SR repo. I don't know the best place to check but this seems to be up to date
https://developer.android.com/tools/releases/platforms

example/android/gradle/wrapper/gradle-wrapper.properties
```diff
@@ -3,4 +3,4 @@ distributionBase=GRADLE_USER_HOME
 distributionPath=wrapper/dists
 zipStoreBase=GRADLE_USER_HOME
 zipStorePath=wrapper/dists
-distributionUrl=https\://services.gradle.org/distributions/gradle-6.7.1-bin.zip
+distributionUrl=https\://services.gradle.org/distributions/gradle-7.5-bin.zip
```
justin-fiedler:
```suggestion
distributionUrl=https\://services.gradle.org/distributions/gradle-8.2.2-bin.zip
```

example/android/app/build.gradle
```diff
@@ -21,12 +21,13 @@ if (flutterVersionName == null) {
...
 }
 
+apply plugin: 'idea'
 apply plugin: 'com.android.application'
 apply plugin: 'kotlin-android'
 apply from: "$flutterRoot/packages/flutter_tools/gradle/flutter.gradle"
 
 android {
-    compileSdkVersion 28
+    compileSdkVersion 33
```
justin-fiedler: Make this 34 too?

example/android/gradle/wrapper/gradle-wrapper.properties
Mercy811: Updated in https://github.com/amplitude/Amplitude-Flutter/pull/171/commits/1a3d02c98c31522b0dab94ea4cf6ce6d1e8f75f1.

example/android/app/build.gradle
Mercy811: Updated in https://github.com/amplitude/Amplitude-Flutter/pull/171/commits/1a3d02c98c31522b0dab94ea4cf6ce6d1e8f75f1.

PR #175, opened 2024-03-01
chore: set library by plugin

Mercy811:
Revert changes in https://github.com/amplitude/Amplitude-Flutter/pull/174. 

Only adding a setting library plugin can update all events including default events' library. 

1. Change PUBLISH_VERSION in `release.config.js` to the latest semantic version
2. `build.gradle` will then generate `BuildConfig.java`
3. Set library in `FlutterLibraryPlugin`

Example `BuildConfig.java`
```
/**
 * Automatically generated file. DO NOT MODIFY
 */
package com.amplitude.amplitude_flutter;

public final class BuildConfig {
  public static final boolean DEBUG = Boolean.parseBoolean("true");
  public static final String LIBRARY_PACKAGE_NAME = "com.amplitude.amplitude_flutter";
  public static final String BUILD_TYPE = "debug";
  // Field from default config.
  public static final String SDK_VERSION = "4.0.0-beta.0";
}

```

Events include default events are set to `amplitude-flutter/4.0.0-beta.0`
https://app.amplitude.com/analytics/amplitude/project/471621/search/amplitude_id%3D842311306390?sessionHandle=Y38ZXvS_Y38ZXvS_MQdqoyW_BzJF_6ebbafb4-dc1d-4bd7-8ad7-0ff29dceed93R&eventId=98a9598b-1f8f-42a2-924a-20cfcb268b77

PR #174, opened 2024-03-01
chore: set library

Mercy811:
Set correct library. `packageVersion` in `lib/constants.dart` will be updated to the next semantic version [here](https://github.com/amplitude/Amplitude-Flutter/blob/9a8f1ce3d28a328916dc14746e54328df9bed7a5/release.config.js#L41) in release process. 

Example event with library `amplitude-flutter/4.0.0-beta.0`
https://app.amplitude.com/analytics/amplitude/project/471621/search/amplitude_id%3D841571714990?sessionHandle=Y33KhyN_Y33K%2Ft4_MPxlUeu_BzJF_4836ed76-622b-4760-a488-b3028363b5c0R&eventId=8ea03829-2e3d-4aa4-baf9-e9591577092c

PR #176, opened 2024-03-06
refactor: update iOS bridge

Mercy811:
- Update iOS bridge file at `ios/Classes/SwiftAmplitudeFlutterPlugin.swift`
- App lifecycle events are not support in the pre-release version. Will have another ticket for it
- Add `swiftlint` to CI
- Don't have iOS native unit tests right now because the [one](https://github.com/amplitude/Amplitude-Flutter/blob/v4.x/android/src/test/kotlin/com/amplitude/amplitude_flutter/AmplitudeFlutterPluginTest.kt) for Android should cover all cases where method channel goes wrong.
- Update set library logics by https://github.com/amplitude/Amplitude-Flutter/pull/176#discussion_r1514798327
  - Before
    - `release.config.js` has to change package version for Flutter, Android (BuildConfig.SDK_VERSION`) and Swift plugins separately
    - Android and Swift library plugins set library for every events
  - Now
    - `release.config.js` only need to change package version for Flutter
    - Flutter pass `library` along with configuration when `init`
    - Native sides read library and pass to library plugins

### DIFF
--- a/android/src/main/kotlin/com/amplitude/amplitude_flutter/AmplitudeFlutterPlugin.kt
+++ b/android/src/main/kotlin/com/amplitude/amplitude_flutter/AmplitudeFlutterPlugin.kt
@@ -28,7 +28,6 @@ class AmplitudeFlutterPlugin : FlutterPlugin, MethodCallHandler, ActivityAware {
 
     private lateinit var channel: MethodChannel
 
-
     companion object {
         private const val methodChannelName = "amplitude_flutter"
     }
@@ -66,7 +65,7 @@ class AmplitudeFlutterPlugin : FlutterPlugin, MethodCallHandler, ActivityAware {
                 amplitude = Amplitude(configuration)
 
                 // Set library
-                amplitude.add(FlutterLibraryPlugin())
+                amplitude.add(FlutterLibraryPlugin(call.argument<String>("library")!!))
 
                 call.argument<String>("logLevel")?.let {
                     amplitude.logger.logMode = Logger.LogMode.valueOf(it.uppercase())
--- a/android/src/main/kotlin/com/amplitude/amplitude_flutter/AmplitudeFlutterPlugin.kt
+++ b/android/src/main/kotlin/com/amplitude/amplitude_flutter/AmplitudeFlutterPlugin.kt
@@ -65,7 +65,11 @@ class AmplitudeFlutterPlugin : FlutterPlugin, MethodCallHandler, ActivityAware {
                 amplitude = Amplitude(configuration)
 
                 // Set library
-                amplitude.add(FlutterLibraryPlugin(call.argument<String>("library")!!))
+                amplitude.add(
+                    FlutterLibraryPlugin(
+                        call.argument<String>("library") ?: "amplitude-flutter/unknown"
+                    )
+                )
 
                 call.argument<String>("logLevel")?.let {
                     amplitude.logger.logMode = Logger.LogMode.valueOf(it.uppercase())
--- a/android/src/main/kotlin/com/amplitude/amplitude_flutter/AmplitudeFlutterPlugin.kt
+++ b/android/src/main/kotlin/com/amplitude/amplitude_flutter/AmplitudeFlutterPlugin.kt
@@ -196,6 +196,7 @@ class AmplitudeFlutterPlugin : FlutterPlugin, MethodCallHandler, ActivityAware {
         call.argument<Boolean>("useBatch")?.let { configuration.useBatch = it }
         call.argument<String>("serverZone")
             ?.let { configuration.serverZone = com.amplitude.core.ServerZone.valueOf(it.uppercase()) }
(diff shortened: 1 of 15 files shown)


ios/Classes/SwiftAmplitudeFlutterPlugin.swift
```diff
@@ -1,252 +1,400 @@
...
+
+            do {
+                let event = try getEvent(args: args)
+                amplitude?.track(event: event)
+                amplitude?.logger?.debug(message: "Track revenue event: \(String(describing: call.arguments))")
+
+                result("revenue called..")
+            } catch {
+                amplitude?.logger?.warn(message: "revenue called but failed.")
+            }
```
justin-fiedler: It looks like these event based sections are all same expect for logging. Could these maybe be consolidated into a single handler to make maintenance easier going forward?

```
case "track":
case "identify":
case "groupIdentify":
case "setGroup":
case "revenue":
   ...handler code
```

ios/Classes/SwiftAmplitudeFlutterPlugin.swift
```diff
@@ -1,252 +1,400 @@
...
+
+            do {
+                amplitude = Amplitude(configuration: try getConfiguration(args: args))
+            } catch {
+                print("Initialization failed.")
+            }
+
+            // Set library
+            amplitude?.add(plugin: FlutterLibraryPlugin())
```
justin-fiedler: nit: instead of maintaining the SDK version in 3 places (flutter, android, swift). Maybe we could pass in the SDK version from Flutter in the call args here.

Semantic release makes it easy to update the version in multiple locations, but still feel a single source of truth would be nice.

justin-fiedler (approved):
Overall looks good. See nits. Thanks @Mercy811 !

ios/Classes/SwiftAmplitudeFlutterPlugin.swift
Mercy811: I tried it at https://github.com/amplitude/Amplitude-Flutter/pull/174 and later found that default events will still have native side library so I reverted the change and started using plugin to work this around at https://github.com/amplitude/Amplitude-Flutter/pull/175.

justin-fiedler: I think this would allow the native library to get the library/version from Flutter
```suggestion
            amplitude?.add(plugin: FlutterLibraryPlugin(library = args.library))
            // or
            // amplitude?.add(plugin: FlutterLibraryPlugin(sdkVersion = args.sdkVersion))
```

ios/Classes/SwiftAmplitudeFlutterPlugin.swift
Mercy811: Consolidate both Swift and Kotlin plugins at https://github.com/amplitude/Amplitude-Flutter/pull/176/commits/d0e5f078b78db6be343696b4de591d0cec2826d8

ios/Classes/SwiftAmplitudeFlutterPlugin.swift
Mercy811: Good idea! Updated at https://github.com/amplitude/Amplitude-Flutter/pull/176/commits/f7029acb36eb24e5dc357dc07fc2eeb786077cc2 
- Pass `library` from Flutter via method channel at `init` call
- Pass to library plugins on native sides
- Set library by library plugins

izaaz (approved):

android/src/main/kotlin/com/amplitude/amplitude_flutter/AmplitudeFlutterPlugin.kt
```diff
@@ -66,7 +65,7 @@ class AmplitudeFlutterPlugin : FlutterPlugin, MethodCallHandler, ActivityAware {
                 amplitude = Amplitude(configuration)
 
                 // Set library
-                amplitude.add(FlutterLibraryPlugin())
+                amplitude.add(FlutterLibraryPlugin(call.argument<String>("library")!!))
```
justin-fiedler: nit: `!!` could maybe throw an error? what about this instead?
```suggestion
                amplitude.add(FlutterLibraryPlugin(call.argument<String>("library") ?? "amplitude-flutter/unknown"))
```

ios/Classes/SwiftAmplitudeFlutterPlugin.swift
justin-fiedler: Awesome thanks!

android/src/main/kotlin/com/amplitude/amplitude_flutter/AmplitudeFlutterPlugin.kt
Mercy811: I like it! Update both Kotlin and Swift at https://github.com/amplitude/Amplitude-Flutter/pull/176/commits/4eb9b6fa273ba26d7fe7d69ce1cb29aac5e33037

PR #179, opened 2024-03-07
feat: update to the latests native SDKs

Mercy811:
- Enable flutter test when on release
- Fix some nits at Flutter at `lib/xxx.dart`
- Enable pre-release on beta branch at `release.config.js` which is triggered when the Release (release.yml) github actions is triggered

After this PR is merged to `v4.x`, do the following to pre-release
1. Create a `beta` branch based on `v4.x`
2. Manually run github actions Release with a dry run to make sure the correct semantic version will be set
3. Manually run github actions Release without a dry run to pre-release. This is trigger `publish.yml` to publish the package to pub.dev 
Here is the [test](https://amplitude.atlassian.net/browse/AMP-94599?focusedCommentId=242548) I ran on the forked repo.

Note that it's not as flexible as [feature-branch-prerelease.yml](https://github.com/amplitude/Amplitude-TypeScript/blob/main/.github/workflows/feature-branch-prerelease.yml) of Browser SDK as it does not support selecting the prerelease branch but fix to beta branch due to the limitation of [semantic-release](https://semantic-release.gitbook.io/semantic-release/) while Browser SDK uses [lerna](https://github.com/amplitude/Amplitude-TypeScript/blob/e8216bc65399106214cc5c0ae4d7e8d053029af4/.github/workflows/publish-v2.yml#L88-L95).

### DIFF
--- a/lib/amplitude.dart
+++ b/lib/amplitude.dart
@@ -12,26 +12,36 @@ import 'package:amplitude_flutter/events/group_identify_event.dart';
 class Amplitude {
   Configuration configuration;
   MethodChannel _channel = const MethodChannel("amplitude_flutter");
+  /// Whether the Amplitude instance has been successfully initialized
+  ///
+  /// ```
+  /// var amplitude = Amplitude(Configuration(apiKey: "apiKey"));
+  /// // If care about init complete
+  /// await amplitude.isBuilt;
+  /// ```
+  late Future<bool> isBuilt;
 
   /// Returns an Amplitude instance
   ///
-  /// Call `init()` to initialize underlying SDKs on native platforms
   /// ```
   /// var amplitude = Amplitude(Configuration(apiKey: "apiKey"));
-  /// await amplitude.init();
+  /// // If care about init complete
+  /// await amplitude.isBuilt;
   /// ```
-  Amplitude(this.configuration);
-
-  /// Initializes an Amplitude instance
-  ///
-  /// ```
-  /// var amplitude = Amplitude(Configuration(apiKey: "apiKey"));
-  /// await amplitude.init();
-  /// ```
-  Future<void> init([MethodChannel? methodChannel]) async {
+  Amplitude(this.configuration, [MethodChannel? methodChannel]){
     _channel = methodChannel ?? this._channel;
-    return await _channel.invokeMethod(
-        "init", this.configuration.toMap());
+    isBuilt = _init();
+  }
+
+  /// Private method to initialize and return a Future<bool>
+  Future<bool> _init() async {
+    try {
+      await _channel.invokeMethod("init", configuration.toMap());
+      return true; // Initialization successful
+    } catch (e) {
+      print("Error initializing Amplitude: $e");
+      return false; // Initialization failed
+    }
   }
 
   /// Tracks an event. Events are saved locally.
--- a/lib/amplitude.dart
+++ b/lib/amplitude.dart
@@ -14,6 +14,12 @@ class Amplitude {
   MethodChannel _channel = const MethodChannel("amplitude_flutter");
 
   /// Returns an Amplitude instance
+  ///
+  /// Call `init()` to initialize underlying SDKs on native platforms
+  /// ```
+  /// var amplitude = Amplitude(Configuration(apiKey: "apiKey"));
+  /// await amplitude.init();
+  /// ```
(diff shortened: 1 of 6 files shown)


izaaz (approved):

release.config.js
```diff
@@ -1,6 +1,7 @@
 module.exports = {
   "branches": [
-    "main"
+    "main",
+    { "name": "beta", "prerelease": true },
```
yuhao900914, line 4: Question:
Will every release be a beta release? How do we graduate the beta release without manually updating this file?

lib/amplitude.dart
```diff
@@ -14,6 +14,12 @@ class Amplitude {
   MethodChannel _channel = const MethodChannel("amplitude_flutter");
 
   /// Returns an Amplitude instance
+  ///
+  /// Call `init()` to initialize underlying SDKs on native platforms
+  /// ```
+  /// var amplitude = Amplitude(Configuration(apiKey: "apiKey"));
+  /// await amplitude.init();
```
justin-fiedler: I'm comparing this to the other SDKs and  think this is a little odd to have 2 different initializations. 

If we start the async `init()` logic in the constructor we could use something like `isBuilt` in Kotlin to await the initialization when needed. wdyt?

```
// dont' care about waiting, runs init in background
var amplitude = Amplitude(Configuration(apiKey: "apiKey"));

// If we care about init complete they can await isBuilt
await amplitude.isBuilt();
val deviceId = amplitude.getDeviceId()
```

release.config.js
Mercy811: Just like Kotlin CI, every release on the beta branch will be a beta release. To graduate the beta release, we need to merge the beta branch to the main branch.

lib/amplitude.dart
Mercy811: Thanks Justin, good call out. I updated at https://github.com/amplitude/Amplitude-Flutter/pull/179/commits/53713fff445ef14e73fa8d7246fdd34415218f19

yuhao900914 (approved):

lib/amplitude.dart
```diff
@@ -12,20 +12,36 @@ import 'package:amplitude_flutter/events/group_identify_event.dart';
...
-  /// await amplitude.init();
+  /// // If care about init complete
+  /// await amplitude.isBuilt;
   /// ```
-  Future<void> init([MethodChannel? methodChannel]) async {
+  Amplitude(this.configuration, [MethodChannel? methodChannel]){
     _channel = methodChannel ?? this._channel;
-    return await _channel.invokeMethod(
-        "init", this.configuration.toMap());
+    isBuilt = _init();
```
justin-fiedler, line 33: Is `isBuilt` idempotent or will it rerun init each time?

lib/amplitude.dart
```diff
@@ -12,20 +12,36 @@ import 'package:amplitude_flutter/events/group_identify_event.dart';
...
   Configuration configuration;
   MethodChannel _channel = const MethodChannel("amplitude_flutter");
+  /// Whether the Amplitude instance has been successfully initialized
+  ///
+  /// ```
+  /// var amplitude = Amplitude(Configuration(apiKey: "apiKey"));
+  /// // If care about init complete
+  /// await amplitude.isBuilt;
+  /// ```
+  late Future<bool> isBuilt;
```
justin-fiedler, line 22: wdyt about returning the Amplitude instance on success instead of true? IT might make for easier usage in the cause customers want to wait e.g.

```
analytics = await Amplitude(...).isBuilt
```

lib/amplitude.dart
Mercy811: `isBuilt` is `Future<Amplitude>` type. It will get the result of `_init()`. `_init()` will only run once. Calling `isBuilt` multiple times will just read the value but not rerun `_init()`.

lib/amplitude.dart
Mercy811: I'd rather keep it as `Future<bool>` instead of `Future<Amplitude?>` to avoid `?`. Customers can always achieve the goal (check whether the instance is initialized or not) by separating the call.

justin-fiedler: Good call on the `?`. Current implementation makes sense.

justin-fiedler (approved):

PR #192, opened 2024-06-21
fix: Adding underlying library version to library field

izaaz:
Setting min iOS SDK to 1.6 and appending underlying SDK library to library version

### DIFF
--- a/android/src/main/kotlin/com/amplitude/amplitude_flutter/FlutterLibraryPlugin.kt
+++ b/android/src/main/kotlin/com/amplitude/amplitude_flutter/FlutterLibraryPlugin.kt
@@ -9,7 +9,11 @@ class FlutterLibraryPlugin(val library: String): Plugin {
     override lateinit var amplitude: Amplitude
 
     override fun execute(event: BaseEvent): BaseEvent? {
-        event.library = library
+        if (event.library == null) {
+            event.library = library
+        } else {
+            event.library = "$library_${event.library}"
+        }
         return super.execute(event)
     }
 }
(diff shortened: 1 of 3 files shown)


crleona (approved):

android/src/main/kotlin/com/amplitude/amplitude_flutter/FlutterLibraryPlugin.kt
```diff
@@ -9,7 +9,11 @@ class FlutterLibraryPlugin(val library: String): Plugin {
     override lateinit var amplitude: Amplitude
 
     override fun execute(event: BaseEvent): BaseEvent? {
-        event.library = library
+        if (event.library == null) {
```
Mercy811, line 12: Shouldn't here be `!= null`?

izaaz: We can only merge the `event.library` value if its not null. If its null, we just want to use the flutter library version `Amplitude-Flutter/4.0.0`.

Mercy811: It seems to be redundant as the library parameter is a string type value in the primary constructor and is declared as a non-nullable type (String), it cannot be null.

Mercy811: `event.library` can be null. I was thinking about `library` of the plugin class.

PR #210, opened 2024-11-27
fix: add swift version in podspec

martin-headspace:
# What's Changed

Added the `swift_version` to the Podspec to help developers with an `add-to-app` setup not run into the missing `SWIFT_VERSION` error when compiling.

# Rationale
When using this plugin on a project that is being compiled for the add-to-app approach in Flutter, developers will usually get this error:

```  
 - amplitude_flutter does not specify a Swift version and none of the targets (Runner) integrating it have the SWIFT_VERSION attribute set. Please contact the author or set the SWIFT_VERSION attribute in at least one of the targets that integrate this pod.
```

To fix this, developers have to go into their generated Podspec and add the `SWIFT_VERSION` environment variable for the build to work. This is not an ideal workflow.

By adding the Swift Version to the Podspec, builds compile without a hitch.

### DIFF
--- a/ios/amplitude_flutter.podspec
+++ b/ios/amplitude_flutter.podspec
@@ -13,6 +13,7 @@ Pod::Spec.new do |s|
   s.public_header_files = 'Classes/**/*.h'
   s.dependency 'Flutter'
   s.dependency 'Amplitude', '8.18.0'
+  s.swift_version    = '4.1'
 
   s.ios.deployment_target = '10.0'
 end


ios/amplitude_flutter.podspec
```diff
@@ -13,6 +13,7 @@ Pod::Spec.new do |s|
   s.public_header_files = 'Classes/**/*.h'
   s.dependency 'Flutter'
   s.dependency 'Amplitude', '8.18.0'
+  s.swift_version    = '5.0'
```
chungdaniel: Our Flutter 3.x SDK points to our old iOS maintenance SDK, which [currently uses '4.2'](https://github.com/amplitude/Amplitude-iOS/blob/main/Amplitude.podspec)

chungdaniel (changes requested):
Thanks @martin-headspace for adding this! We should also change the commit to be `fix:` rather than `chore:` so that it'll trigger a build/release.

Some extra context: We're working on releasing a 4.x of our Flutter SDK which will point to our iOS Swift SDK, which will use a swift_version of '5.9'. I've made a note to add this change to our beta branch for including a swift_version for 4.x

ios/amplitude_flutter.podspec
martin-headspace: Thank you so much for looking at this @chungdaniel ! 
I've now modified the title of the commit and made the `swift_version` 4.2

ios/amplitude_flutter.podspec
```diff
@@ -13,6 +13,7 @@ Pod::Spec.new do |s|
   s.public_header_files = 'Classes/**/*.h'
   s.dependency 'Flutter'
   s.dependency 'Amplitude', '8.18.0'
+  s.swift_version    = '4.2'
```
chungdaniel: Sorry I meant `'4.1'`! Other than that looks good. I need to double check for typos going forward.

chungdaniel (changes requested):

ios/amplitude_flutter.podspec
martin-headspace: No worries, it happens! Changed now ❤️

chungdaniel (approved):

Issue #211, opened 2024-11-27
(iOS)(add-to-app): SWIFT_VERSION missing when compiling the iOS app


martin-headspace:
## Expected Behavior

On an add-to-app build, the app using this plugin should compile successfully and without a hitch.

## Current Behavior

Developers get the following error:

`    - amplitude_flutter does not specify a Swift version and none of the targets (Runner) integrating it have the SWIFT_VERSION attribute set. Please contact the author or set the SWIFT_VERSION attribute in at least one of the targets that integrate this pod.
`

Developers _can_ fix this by setting the SWIFT_VERSION environment variable to Swift 5.0 or above.

`ENV['SWIFT_VERSION'] = '5'`

Which, on most Flutter apps is a one time fix, but add-to-app developers have to either set this environment variable on their generated projects every time they compile OR they have to set it elsewhere that can be accessed.

## Possible Solution
#210 

## Steps to Reproduce
 

1. Create or reuse an existing add-to-app Flutter module.
2. Try to compile for an iOS Device.
3. Voila

## Environment
- SDK Version: Latest
- Flutter Version: Latest Stable
- Device: All iOS
- Device OS and Version: Any iOS

chungdaniel:
Fixed with #210

chungdaniel:
Actually will leave open until build goes out :)

chungdaniel:
@martin-headspace 3.16.6 including the fix has been released!

PR #217, opened 2025-01-14
feat: add support for macOS

chungdaniel:
Primary changes:
- Move iOS code `sharedDarwinSource` to reuse for macOS according to https://docs.flutter.dev/packages-and-plugins/developing-packages#shared-ios-and-macos-implementations - thanks to @jointhejourney for their work in #156 for the idea
- Added support for macOS in example app

Extra:
- Added swift_version to `amplitude_flutter.podspec`, follow-up to #211 for v4
- Bumped flutter depdendency to 3.7.0+ to support `sharedDarwinSource`

Note:
- PR should probably be focused mostly on non-example files

Addresses #46 
Jira: [[SDK - Flutter] Flutter macOS Support](https://amplitude.atlassian.net/browse/AMP-35292)


README.md
```diff
@@ -22,9 +22,9 @@ From Amplitude Flutter v4, we bump up the kotlin version to v1.9.22 to support l
 
 The following matrix lists the minimum support for Amplitude Flutter SDK version.
 
-| Flutter | Amplitude Flutter | Gradle | Android Gradle Plugin | Kotlin Gradle Plugin |
-|---------|-------------------|-------|-----------------------|-----------------------|
-|`3.3+`   | `4.+`             | `8.2` | `8.2.2`               | `1.9.22`              |
+|  Dart   | Flutter | Amplitude Flutter | Gradle | Android Gradle Plugin | Kotlin Gradle Plugin |
+|---------|---------|-------------------|-------|-----------------------|-----------------------|
+| `3.7.0+`|`3.3+`   | `4.+`             | `8.2` | `8.2.2`               | `1.9.22`              |
```
Mercy811: What requires Dart version? `sharedDarwinSource` requires Flutter 3.7 right?

darwin/amplitude_flutter.podspec
```diff
@@ -0,0 +1,33 @@
...
+  # s.resource_bundles = {'amplitude_flutter_privacy' => ['Resources/PrivacyInfo.xcprivacy']}
+
+  s.ios.dependency 'Flutter'
+  s.osx.dependency 'FlutterMacOS'
+  s.ios.deployment_target = '13.0'
+  s.osx.deployment_target = '10.15'
+
+  s.pod_target_xcconfig = { 'DEFINES_MODULE' => 'YES' }
+  s.swift_version = '5.9'
+  s.dependency 'AmplitudeSwift', '~> 1.6'
```
izaaz: Can we update to the latest version of the SDK as the min version?

darwin/Resources/PrivacyInfo.xcprivacy
```diff
@@ -0,0 +1,14 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
izaaz: Do we need to add this? The underlying SDK also has a privacy manifest which is more up to date.

izaaz (approved):
Mostly LGTM with a few minor comments.

README.md
chungdaniel: Ah, good catch. Yes, Flutter requires 3.7 for `sharedDarwinSource`, but we need dart 3.3+ to support Flutter web after adopting `js_interop`. I'll swap those around

Mercy811 (approved):
Thank you @chungdaniel, LGTM other than flutter & dart version requirements.

darwin/Resources/PrivacyInfo.xcprivacy
chungdaniel: It was added by default when I tried adding macOS as a platform as a result of me updating to latest flutter locally. I think it should be fine to remove it.

PR #226, opened 2025-02-06
chore: Use minor version in podfile syntax for Amplitude Swift SDK

chungdaniel:
Currently the podspec uses patch version with optimistic operator (`~> 1.11.2`) to specify pod version for the Amplitude-Swift SDK, which translates to versions 1.11.2 <= version < 1.12. We would like to instead include up to but excluding the next major version.

Jira: AMP-98098


Mercy811 (approved):
Thanks @chungdaniel for the fix

PR #230, opened 2025-02-28
feat: add getUserId and getDeviceId functions on web/android/ios

chungdaniel:
Flutter SDK 3.x had methods for `getUserId` and `getDeviceId`, while Flutter SDK 4 does not. Customers on the 3.x SDK using `getDeviceId`/`getUserId` are blocked from upgrading to 4.x as a result. All platforms already have `getUserId` and `getDeviceId` so simply add appropriate calls.

Primary changes:
- Add `getUserId` and `getDeviceId` calls to `amplitude.dart`
- Add corresponding calls in Android, iOS, and Web plugin files to call underlying SDK

Jira: https://amplitude.atlassian.net/browse/AMP-125025

### DIFF
--- a/darwin/Classes/SwiftAmplitudeFlutterPlugin.swift
+++ b/darwin/Classes/SwiftAmplitudeFlutterPlugin.swift
@@ -67,6 +67,12 @@ import AmplitudeSwift
                 amplitude?.logger?.warn(message: "\(call.method) called but failed.")
             }
 
+        case "getUserId":
+            let userId = amplitude?.getUserId()
+            amplitude?.logger?.debug(message: "Get user Id: \(String(describing: userId))")
+
+            result(userId)
+
         case "setUserId":
             guard let args = call.arguments as? [String: Any] else {
                 print("\(call.method) called but call.arguments type casting failed.")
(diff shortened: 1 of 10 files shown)


darwin/Classes/SwiftAmplitudeFlutterPlugin.swift
```diff
@@ -67,6 +67,12 @@ import AmplitudeSwift
                 amplitude?.logger?.warn(message: "\(call.method) called but failed.")
             }
 
+        case "getUserId":
+            let userId = amplitude?.getUserId()
+            amplitude?.logger?.debug(message: "Get user Id: \(String(describing: userId))")
```
Mercy811: nit: keep aligned with the android plugin
```suggestion
            amplitude?.logger?.debug(message: "Get userId: \(String(describing: userId))")
```

darwin/Classes/SwiftAmplitudeFlutterPlugin.swift
```diff
@@ -82,6 +88,12 @@ import AmplitudeSwift
 
             result("serUserId called..")
 
+        case "getDeviceId":
+            let deviceId = amplitude?.getDeviceId()
+            amplitude?.logger?.debug(message: "Get device Id: \(String(describing: deviceId))")
```
Mercy811: nit: the same as above
```suggestion
            amplitude?.logger?.debug(message: "Get deviceId: \(String(describing: deviceId))")
```

lib/amplitude_web.dart
```diff
@@ -48,11 +48,27 @@ class AmplitudeFlutterPlugin {
         JSObject event = getEvent(call);
         amplitude.track(event);
       }
+      case "getUserId":
+      {
+        JSString? userId = amplitude.getUserId();
+        if (userId == null) {
+          return null;
+        }
+        return userId.toDart;
```
Mercy811, line 57: nit: can we use null-aware operator here?
```suggestion
        return amplitude.getUserId()?.toDart;
```

Mercy811 (approved):
Thanks @chungdaniel. LGTM. See nits

PR #233, opened 2025-03-05
refactor: Hides configuration from public interface and make isBuilt final

gabrielgarciagava:
**TL;DR** This PR achieves what is described in this link: https://amplitude.com/docs/sdks/analytics/flutter/flutter-sdk-4-0-migration-guide#configuration

> Configuration is immutable after you pass it to Amplitude.

### Description
The purpose of these changes is because having both configuration and isBuilt publicly mutable is misleading, since changing those attributes will have no impact.

For example, currently it is possible to initialize Amplitude 
```
final amplitude = Amplitude(Configuration(apiKey: "xxx"));
```

And then write
```
amplitude.configuration.flushQueueSize = 1;
```
Thinking this would lead to a on-the-fly reconfiguration of amplitude, which is not true.

Similarly but less important, I made isBuilt final since previosuly it would be possible to modify it, which makes no sense.

### DIFF
--- a/lib/amplitude.dart
+++ b/lib/amplitude.dart
@@ -10,7 +10,6 @@ import 'events/base_event.dart';
 import 'events/group_identify_event.dart';
 
 class Amplitude {
-  Configuration configuration;
   MethodChannel _channel = const MethodChannel('amplitude_flutter');
 
   /// Whether the Amplitude instance has been successfully initialized
@@ -20,7 +19,7 @@ class Amplitude {
   /// // If care about init complete
   /// await amplitude.isBuilt;
   /// ```
-  late Future<bool> isBuilt;
+  late final Future<bool> isBuilt;
 
   /// Returns an Amplitude instance
   ///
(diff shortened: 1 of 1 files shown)


gabrielgarciagava:
This removes a field from the public interface, which means it is a breaking change.
However, using this field is completely useless. If someone is using it, it is probably just a bug.

So my question is if this should create a new major release or not.

Mercy811:
Hi @gabrielgarciagava, thanks for bringing this up. It makes sense to me that it's kind of misleading. We will make `isBuilt` final but will keep `configuration` there to avoid breaking changes.

gabrielgarciagava:
Of course. But please consider hiding it in a next major release. Or at least making the field final, such as all configuration fields as well.

This way a user of the library will never be able to modify the configuration anymore after amplitude is initialized, which makes sense since it is not supported.

gabrielgarciagava:
Just one use case to show you the motivation behind it.

When using version 3 we used to have the following code
```
final amplitude = Amplitude();
...
amplitude.enableCoppaControl();
```

When moving to version 4, someone just followed the naive substitution, which was allowed by the library interface
```
final amplitude = Amplitude(Configuration(apiKey: ...));
...
amplitude.configuration.enableCoppaControl = true;
```

Which of course is now wrong, but one needs to read the library implementation to realize it.

lib/amplitude.dart
```diff
@@ -29,13 +28,13 @@ class Amplitude {
   /// // If care about init complete
   /// await amplitude.isBuilt;
   /// ```
-  Amplitude(this.configuration, [MethodChannel? methodChannel]) {
+  Amplitude(Configuration configuration, [MethodChannel? methodChannel]) {
```
chungdaniel, line 31: As discussed, if you could revert the changes to hiding `Configuration` for now and leave in making `isBuilt` final I'd be happy to approve this and get this merged!

chungdaniel (changes requested):

PR #237, opened 2025-03-13
feat: add getSessionId for all platforms

chungdaniel:
`getSessionId` was not supported on Flutter SDK 4, while it was supported in 3.x. This PR adds that capability.

Jira: AMP-126150

### DIFF
--- a/android/src/main/kotlin/com/amplitude/amplitude_flutter/AmplitudeFlutterPlugin.kt
+++ b/android/src/main/kotlin/com/amplitude/amplitude_flutter/AmplitudeFlutterPlugin.kt
@@ -122,6 +122,13 @@ class AmplitudeFlutterPlugin : FlutterPlugin, MethodCallHandler, ActivityAware {
                 result.success("setDeviceId called..")
             }
 
+            "getSessionId" -> {
+                val sessionId = amplitude.sessionId
+                amplitude.logger.debug("Get sessionId: $sessionId")
+
+                result.success(sessionId)
+            }
+
             "reset" -> {
                 amplitude.reset()
                 amplitude.logger.debug("Reset userId and deviceId.")
--- a/darwin/Classes/SwiftAmplitudeFlutterPlugin.swift
+++ b/darwin/Classes/SwiftAmplitudeFlutterPlugin.swift
@@ -118,6 +118,12 @@ import AmplitudeSwift
 
             result("setDeviceId called..")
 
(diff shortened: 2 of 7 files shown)


android/src/main/kotlin/com/amplitude/amplitude_flutter/AmplitudeFlutterPlugin.kt
```diff
@@ -122,6 +122,13 @@ class AmplitudeFlutterPlugin : FlutterPlugin, MethodCallHandler, ActivityAware {
                 result.success("setDeviceId called..")
             }
 
+            "getSessionId" -> {
+                val sessionId = amplitude.sessionId
```
Mercy811, line 126: looks like we should also call `amplitude.getSessionId()` in Android SDK?

chungdaniel: Hey @Mercy811 I don't think that is a valid method in Android SDK. The Amplitude-Kotlin SDK docs mentions it, though I think they are wrong. Looking at the code, I only see getters but no `getSessionId` method defined.
See here: https://github.com/amplitude/Amplitude-Kotlin/blob/main/android/src/main/java/com/amplitude/android/Amplitude.kt#L27
No mention of `getSessionId` (or `sessionId`) in core either: https://github.com/amplitude/Amplitude-Kotlin/blob/main/core/src/main/java/com/amplitude/core/Amplitude.kt

Mercy811 (approved):

android/src/main/kotlin/com/amplitude/amplitude_flutter/AmplitudeFlutterPlugin.kt
Mercy811: @igor-amp please review if Android SDK interface is used correctly here.

igor-amp: ic that SessionID only exists on Android and is tightly connected to Android version of timeline. As of missing getSessionId() functions: Kotlin introduces properties that encapsulate fields while providing getter and setter methods implicitly. Meaning that you can't have properties and same name functions in the same class. It will not build. So this use is acceptable.

igor-amp (approved):
Approve for Android.

PR #241, opened 2025-03-22
feat: add multiple instances support

### LINKED ISSUE #239: Can't run two Amplitude instances
I am trying to implement two Amplitude configurations that will run in parallel, following the [sdk v4 guide](https://amplitude.com/docs/sdks/analytics/flutter/flutter-sdk-4#initialize-the-sdk).

This is how I set it up:

```dart
final amp1 = Amplitude(Configuration(
  apiKey: apiKey,
  defaultTracking: const DefaultTrackingOptions(sessions: false),
));
await amp1.isBuilt;

final amp2 = Amplitude(Configuration(
  instanceName: 'secondary',
  apiKey: apiKey,
  serverUrl: serverUrl,
  defaultTracking: const DefaultTrackingOptions(sessions: true),
));
await amp2.isBuilt;
```
This will be used to send some events to regular amplitude server, and other events to another server.

I tested it on iOS and ~it seems to be ok~ [update: it doesn't work either]. ~I will test on Android later~ [update: same bug on Android]. But my issue is on Web. It seems like there can be only 1 instance, and I can't make them work independently.

I added the JS code to `web/index.html` as required:
```html
<script type="text/javascript">
    !function(){"use strict";!function(e,t){var r=e.amplitude||{_q:[],_iq:{}}
    ;if(r.invoked)e.console&&console.error&&console.error("Amplitude snippet has been loaded.")
    ;else{var n=function(e,t){e.prototype[t]=function(){return this._q.push({name:t,args:Array.prototype.slice.call(arguments,0)}),this}}
    ,s=function(e,t,r){return function(n){e._q.push({name:t,args:Array.prototype.slice.call(r,0),resolve:n})}}
    ,o=function(e,t,r){e._q.push({name:t,args:Array.prototype.slice.call(r,0)})}
    ,i=function(e,t,r){e[t]=function(){if(r)return{promise:new Promise(s(e,t,Array.prototype.slice.call(arguments)))};o(e,t,Array.prototype.slice.call(arguments))}}
    ,a=function(e){for(var t=0;t<g.length;t++)i(e,g[t],!1);for(var r=0;r<m.length;r++)i(e,m[r],!0)}
    ;r.invoked=!0;var c=t.createElement("script")
    ;c.type="text/javascript",c.integrity="sha384-R0H1kXlk6r2aEQMtwVcPolpk0NAuIqM/8NlxAv24Gr3/PBJPl+9elu0bc3o/FDjR",c.crossOrigin="anonymous",c.async=!0
    ,c.src="https://cdn.amplitude.com/libs/analytics-browser-2.11.10-min.js.gz"
    ,c.onload=function(){e.amplitude.runQueuedFunctions||console.log("[Amplitude] Error: could not load SDK")}
    ;var l=t.getElementsByTagName("script")[0];l.parentNode.insertBefore(c,l);for(var u=function(){return this._q=[],this}
    ,p=["add","append","clearAll","prepend","set","setOnce","unset","preInsert","postInsert","remove","getUserProperties"],d=0;d<p.length;d++)n(u,p[d]);r.Identify=u
    ;for(var f=function(){return this._q=[],this},v=["getEventProperties","setProductId","setQuantity","setPrice","setRevenue","setRevenueType","setEventProperties"],y=0;y<v.length;y++)n(f,v[y]);r.Revenue=f;v ar g=["getDeviceId","setDeviceId","getSessionId","setSessionId","getUserId","setUserId","setOptOut","setTransport","reset","extendSession"],m=["init","add","remove","track","logEvent","identify","groupIdentify","setGroup","revenue","flush"]
    ;a(r),r.createInstance=function(e){return r._iq[e]={_q:[]},a(r._iq[e]),r._iq[e]},e.amplitude=r}}(window,document)}();
  </script>
```

The initialization doesn't seem to fail, but I only can get events to the first configuration.
If I add a long enough delay between the initializations:
```dart
await Future.delayed(const Duration(seconds: 2));
```
the it starts sending all events to the second instance.

I was trying to debug the js code, and I came as far as to this piece of code:
Image
> Some part of https://cdn.amplitude.com/libs/analytics-browser-2.11.10-min.js.gz

This `this.initializing` check seems to be preventing the second instance from being initialized properly. That's why I decided to add the delay in between the initializations. It kinda worked, because the second `init` seemed to execute as expected (this far, at least), but then all events would be sent via this second config. 

## Expected Behavior
I expect each client to behave and work independently, for any platform

## Current Behavior
When on Web, all the events are sent using the same Amplitude instance
~On iOS it seems to be working as intended~ [update: same thing as on web]
~I haven't tested on Android yet~ [update: same thing as on web and iOS]

Image

## Possible Solution

Not a clue 😕

## Steps to Reproduce
1. Create an Amplitude instance with a given configuration
2. Create a second Amplitude instance with a different configuration
3. Track an event with each instance
4. Verify which configuration was used for each event

## Environment
- SDK Version: 4.1.0
- Flutter Version: 3.24.3 • channel stable
- Device: Chrome, iOS Simulator
- Device OS and Version: macOS, iOS 18.2, Android 15

guilhermearaujo:
# Add Multi-Instance Support to Amplitude Flutter Plugin

This PR adds support for multiple Amplitude instances in the Flutter plugin, allowing developers to initialize and manage multiple Amplitude instances with different configurations within the same app.

### Core Flutter Implementation
- Modified `Amplitude` class to pass instance name to all channel invocations
- Updated method channel calls to include instance name in the arguments
- Added instance name to all event tracking and configuration methods

#### Android
- Added a map to store multiple Amplitude instances
- Updated method channel handler to support instance-specific operations
- Modified event handling to work with instance-specific configurations
- Updated tests to include instance name in method calls

#### iOS/macOS
- Replaced single Amplitude instance with a dictionary of instances
- Updated method channel handler to support instance-specific operations
- Modified event handling to work with instance-specific configurations

#### Web
- Added JS interop call to `createInstance`
- Added support for multiple Amplitude instances in the JavaScript bridge
- Updated method channel handler to support instance-specific operations
- Modified event handling to work with instance-specific configurations

## Testing
- Updated all platform-specific tests to include instance name in method calls
- Added test coverage for instance-specific operations
- Verified backward compatibility with existing single-instance usage

## Breaking Changes
None. The changes maintain backward compatibility with existing single-instance usage by defaulting to `$default_instance` when no instance name is provided.

## Example Usage
```dart
// Initialize multiple instances
final instance1 = Amplitude(Configuration(instanceName: 'instance1', apiKey: 'apiKey_one'));
final instance2 = Amplitude(Configuration(instanceName: 'instance2', apiKey: 'apiKey_two'));

// Track events to specific instances
instance1.track(event1);
instance2.track(event2);
```

This closes #239


lib/amplitude_web.dart
```diff
@@ -27,6 +29,30 @@ class AmplitudeFlutterPlugin {
...
+      String apiKey = args['apiKey'];
+      JSObject configuration = getConfiguration(call);
+
+      // Set library
+      Amplitude instance = amplitude.createInstance();
+      instance.add(createJSInteropWrapper(FlutterLibraryPlugin(
+          args['library'] ?? 'amplitude_flutter/unknown')));
+      instance.init(apiKey, configuration);
+
+      instances[args['instanceName'] ?? '\$default_instance'] = instance;
```
chungdaniel: nit: use `Constants.defaultInstanceName`

lib/amplitude_web.dart
```diff
@@ -27,6 +29,30 @@ class AmplitudeFlutterPlugin {
...
+    Amplitude? instance =
+        instances[call.arguments['instanceName'] ?? '\$default_instance'];
+
+    if (instance == null) {
+      throw Exception(
+          'instance not found: ${call.arguments['instanceName'] ?? '\$default_instance'}');
+    }
+
     switch (call.method) {
       case "init":
```
chungdaniel: Should we just remove the `init` block from this switch since it's handled by the `if (call.method == 'init'){` logic you added on L32 that returns early?

chungdaniel (changes requested):
Thanks for looking into this and contributing! Looks good except for two small nits. Let's get those changes in then we can hopefully create a release soon! :)

lib/amplitude_web.dart
guilhermearaujo: Sure! I forgot to clean that up

lib/amplitude_web.dart
guilhermearaujo: Fixed! Sorry, I'm not very familiar with the project 🙄

chungdaniel (approved):

PR #253, opened 2025-04-16
fix: Unnecessary double await call

cfoltran:


lib/amplitude.dart
```diff
@@ -270,7 +270,7 @@ class Amplitude {
 
   /// Flush events in storage.
   Future<void> flush() async {
-    await await _channel
+    await _channel
```
chungdaniel: Could we return this instead?
```suggestion
    return await _channel
```

chungdaniel (changes requested):
@cfoltran good catch! Thanks for catching this. Just one small change

lib/amplitude.dart
cfoltran: @chungdaniel aha good catch too 🤝

chungdaniel (approved):

PR #268, opened 2025-06-26
fix: lock iOS and Android SDK dependencies to specific versions

chungdaniel:
### Description
Amplitude-Flutter was not locked to a particular file for iOS and Android SDKs. As a result, users of the SDK could accidentally pull in beta versions.

### Approach taken
Lock Android and iOS versions to specific versions to prevent this in the future.


crleona:
Let's discuss this. We generally do not want to take this approach.

PR #273, opened 2025-07-10
fix: [android] Remove duplicate install event tracking

chungdaniel:
Remove manual trackAppUpdatedInstalledEvent call in Flutter plugin since Android SDK automatically handles app lifecycle events when appLifecycles is enabled in DefaultTrackingOptions. This prevents duplicate install events that only occurred in Flutter apps. 

Additionally, added deeplink handling for the Android example app.

Tested manually that installing app no longer double tracks install 

Jira: https://amplitude.atlassian.net/browse/AMP-133754

### DIFF
--- a/android/src/main/kotlin/com/amplitude/amplitude_flutter/AmplitudeFlutterPlugin.kt
+++ b/android/src/main/kotlin/com/amplitude/amplitude_flutter/AmplitudeFlutterPlugin.kt
@@ -168,17 +168,6 @@ class AmplitudeFlutterPlugin : FlutterPlugin, MethodCallHandler, ActivityAware {
             } else {
                 val utils = DefaultEventUtils(amplitude)
 
-                if (appLifecycles) {
-                    val packageManager = ctxt.packageManager
-                    val packageInfo = try {
-                        packageManager.getPackageInfo(ctxt.packageName, 0)
-                    } catch (ex: PackageManager.NameNotFoundException) {
-                        println("Error occurred in getting package info. " + ex.message)
-                        null
-                    }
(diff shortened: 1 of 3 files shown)


android/src/main/kotlin/com/amplitude/amplitude_flutter/AmplitudeFlutterPlugin.kt
```diff
@@ -168,17 +168,6 @@ class AmplitudeFlutterPlugin : FlutterPlugin, MethodCallHandler, ActivityAware {
...
 
-                if (appLifecycles) {
-                    val packageManager = ctxt.packageManager
-                    val packageInfo = try {
-                        packageManager.getPackageInfo(ctxt.packageName, 0)
-                    } catch (ex: PackageManager.NameNotFoundException) {
-                        println("Error occurred in getting package info. " + ex.message)
-                        null
-                    }
-                    packageInfo?.let { utils.trackAppUpdatedInstalledEvent(it) }
```
aliaksandr-kazarez, line 179: should this be removed for other platforms also?
https://github.com/amplitude/Amplitude-Flutter/blob/b41a85acd1e708616b5b1bd0a328c8e8318a6613/darwin/Classes/SwiftAmplitudeFlutterPlugin.swift#L51

aliaksandr-kazarez (approved):

android/src/main/kotlin/com/amplitude/amplitude_flutter/AmplitudeFlutterPlugin.kt
chungdaniel: Good catch! Seems like iOS still tracked app installs with it removed, and there was some deduping occurring on the iOS SDK.

PR #276, opened 2025-08-05
fix: adding ability to track revenue currency

chungdaniel:
Exposes a field for currency in the Flutter SDK to get revenue currency.

Manually tested to see that currency gets sent in events. Matched what was sent by native iOS SDK example app. Works with or without a value provided for currency (gets omitted if not present).

image

Jira: https://amplitude.atlassian.net/browse/AMP-124899


aliaksandr-kazarez (approved):
LGTM 👌

PR #291, opened 2026-03-24
feat(internal): improvements to support other Amplitude Flutter SDKs

Cavallando:
**Context**
- Guides & Surveys / Engagement team have been working to bring these Amplitude SDKs more in line with the SDK plugin system
- Flutter is still a bit behind from this and we worked with the analytics SDK team to figure out how we can enable Guides & Surveys / Engagement to support this
- We were a bit limited because both G&S and Analytics SDKs work as a wrapper around the native instance
- We decided that we could expose a way for other SDK's to get the underlying `Amplitude` instance from the `instances` map to use and hook up in a native context.

**Changes**
- On both iOS and Android, I set up a global private variable to store the instance of the `Plugin` implementation we have. The FlutterEngine will only use a single instance.
- Then I added `getAmplitudeInstanceById` function to actually retrieve this from the global private var
- This will then get used by AmplitudeEngagement when setting up that SDK via plugin (ie. in dart `installAmplitudeEngagementPlugin(amplitude: Amplitude)`
- For more info on how this will be used, see this PR in G&S repo https://github.com/amplitude/amplitude-gs-mobile-sdk/pull/640

---

> [!NOTE]
> **Medium Risk**
> Introduces global plugin singletons on Android/iOS to expose the underlying `Amplitude` instances to other native SDKs, which could impact lifecycle/memory behavior if the engine attachment/detachment is mishandled.
> 
> **Overview**
> Enables other Amplitude Flutter SDKs to hook into the underlying native `Amplitude` instance by exposing a native `getAmplitudeInstanceById(instanceName)` accessor on both Android and Darwin.
> 
> Android stores a global `pluginInstance` for the engine lifecycle and adds an `androidx.annotation` dependency to scope the new API as library-internal. Documentation is updated with internal guidance on integrating additional Flutter SDKs/plugins via the retrieved native instance.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 48a000de2a99b2a7085e743bfda89f5d8fbc1433. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/android/src/main/kotlin/com/amplitude/amplitude_flutter/AmplitudeFlutterPlugin.kt
+++ b/android/src/main/kotlin/com/amplitude/amplitude_flutter/AmplitudeFlutterPlugin.kt
@@ -21,7 +21,12 @@ import io.flutter.embedding.engine.plugins.activity.ActivityAware
 import io.flutter.embedding.engine.plugins.activity.ActivityPluginBinding
 import java.lang.ref.WeakReference
 
+// Global variable to store the plugin instance for the duration of the app's lifecycle
+// This is used to access the plugin instance from a native context in other Amplitude SDKs.
+private var pluginInstance: AmplitudeFlutterPlugin? = null
+
 class AmplitudeFlutterPlugin : FlutterPlugin, MethodCallHandler, ActivityAware {
+    private val instances: MutableMap<String, Amplitude> = mutableMapOf()
     private var activity: WeakReference<Activity?> = WeakReference(null)
     lateinit var ctxt: Context
     private var appOpenedTracked = false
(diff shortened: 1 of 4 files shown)


android/src/main/kotlin/com/amplitude/amplitude_flutter/AmplitudeFlutterPlugin.kt
```diff
@@ -341,3 +361,11 @@ class AmplitudeFlutterPlugin : FlutterPlugin, MethodCallHandler, ActivityAware {
...
     }
 }
+
+@RequiresOptIn(
+    message = "This is an internal Amplitude API and is not intended for use outside of Amplitude SDKs.",
+    level = RequiresOptIn.Level.ERROR
+)
+@Retention(AnnotationRetention.BINARY)
+@Target(AnnotationTarget.FUNCTION)
+annotation class InternalAmplitudeApi
```
Cavallando: On iOS, I was able to use the `@_spi` annotation on the signature / import but on Android, outside of using reflection, it seemed like this could be a way to deter folks from using.

But totally understand if not desired, happy to change to something else!

android/src/main/kotlin/com/amplitude/amplitude_flutter/AmplitudeFlutterPlugin.kt
```diff
@@ -31,6 +35,20 @@ class AmplitudeFlutterPlugin : FlutterPlugin, MethodCallHandler, ActivityAware {
...
+
+        /**
+         * Returns an Amplitude instance by its instance name.
+         * This method is intended to be used by other Amplitude SDKs to be able to
+         * access the underlying Amplitude instance from a native context.
+         *
+         * @param id The instance name of the Amplitude instance.
+         * @return The Amplitude instance or null if not found.
+         */
+        @InternalAmplitudeApi
```
Cavallando: TODO: Remove private stuff here and @_spi in iOS

chungdaniel:
Just a headsup: `chore:` won't trigger a release since semantic versioning only takes into account `feat` or `fix` to bump. May want to change to a feat or just push a dummy commit (in a separate PR or standalone commit) with a generic comment if we don't really want this to appear in our changelogs

chungdaniel (approved):

aliaksandr-kazarez (approved):
LGTM

Issue #283, opened 2025-10-10
No direct method <init>


k0shk0sh:
Can you please use a direct version on the Android SDK rather than version range.  

## Expected Behavior

App not crashing. 

## Current Behavior

App is crashing

## Possible Solution

Use an exact version of the Android SDK

## Steps to Reproduce
 

1. We use amplitude_flutter_ 4.3.7 
2. We also use gradle lockfile to lock our dependencies
3. On installing Android or shipping to users via CLI, the SDK always fetches SDK version that we never TESTED and targeted

## Environment
- SDK Version: 4.3.7
- Flutter Version: 3.35.1
- Device: Pixel 10 PRO
- Device OS and Version: Latest

We get this crash

E/AndroidRuntime(18862): java.lang.NoSuchMethodError: No direct method <init>(Ljava/lang/String;Landroid/content/Context;IILjava/lang/String;ZLcom/amplitude/core/StorageProvider;Lcom/amplitude/core/LoggerProvider;Ljava/lang/Integer;Ljava/lang/String;Lkotlin/jvm/functions/Function3;IZLcom/amplitude/core/ServerZone;Ljava/lang/String;Lcom/amplitude/core/events/Plan;Lcom/amplitude/core/events/IngestionMetadata;ZZZLcom/amplitude/android/TrackingOptions;ZZZJLjava/util/Set;JLcom/amplitude/core/StorageProvider;Lcom/amplitude/id/IdentityStorageProvider;ZLjava/lang/Boolean;Ljava/lang/String;Ljava/lang/Long;Lcom/amplitude/core/utilities/http/HttpClientInterface;Lcom/amplitude/android/InteractionsOptions;IILkotlin/jvm/internal/DefaultConstructorMarker;)V in class Lcom/amplitude/android/Configuration; or its super classes (declaration of 'com.amplitude.android.Configuration' appears in /data/app/~~W9vOLmt4lYo1gNqJWxKDcw==/de.fastic.app.debug-_eZWkr4hD1evihwNxu8OPQ==/base.apk!classes28.dex)

E/AndroidRuntime(18862): at com.amplitude.amplitude_flutter.AmplitudeFlutterPlugin.getConfiguration(AmplitudeFlutterPlugin.kt:191)

E/AndroidRuntime(18862): at com.amplitude.amplitude_flutter.AmplitudeFlutterPlugin.onMethodCall(AmplitudeFlutterPlugin.kt:64)

[...]

PR #295, opened 2026-03-31
fix: update kotlin SDK to 1.27

polbins:
- this should use the config builder which has a safer constructor than config object

---

> [!NOTE]
> **Medium Risk**
> Changes Amplitude Android SDK dependency resolution and the plugin’s initialization/config mapping, which can alter runtime behavior (autocapture flags, batching/session tracking) if the builder semantics differ or a newer 1.27.x release is pulled in.
> 
> **Overview**
> Updates the Android dependency pin from a fixed `com.amplitude:analytics-android:1.25.3` to the `1.27.+` minor line to reduce binary-compatibility issues while allowing patch updates.
> 
> Refactors plugin initialization to build SDK configuration via `ConfigurationBuilder` (instead of mutating a `Configuration` instance), and maps `defaultTracking` into the SDK’s `autocapture` set (`SESSIONS`, `APP_LIFECYCLES`, `DEEP_LINKS`), updating lifecycle/deeplink tracking checks accordingly.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 337e5016e4f07da7281f9fb5e39688f086912682. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/android/build.gradle
+++ b/android/build.gradle
@@ -46,10 +46,10 @@ dependencies {
     implementation "org.jetbrains.kotlin:kotlin-stdlib-jdk7:$kotlin_version"
     implementation "org.jetbrains.kotlin:kotlin-reflect:1.6.10"
 
-    // Locked to specific version to prevent binary compatibility issues.
+    // Locked to specific minor version to prevent binary compatibility issues.
     // Version range [1.22,2.0) previously caused NoSuchMethodError when Gradle
     // resolved different versions at compile-time vs runtime (see #283).
-    implementation 'com.amplitude:analytics-android:1.25.3'
+    implementation 'com.amplitude:analytics-android:1.27.+'
     testImplementation 'org.mockito:mockito-core:4.0.0'
     testImplementation "io.mockk:mockk:1.12.4"
     testImplementation "io.mockk:mockk-agent-jvm:1.11.0"
--- a/android/build.gradle
+++ b/android/build.gradle
@@ -47,8 +47,8 @@ dependencies {
     implementation "org.jetbrains.kotlin:kotlin-reflect:1.6.10"
 
     // Locked to specific minor version to prevent binary compatibility issues.
-    // Version range [1.22,2.0) previously caused NoSuchMethodError when Gradle
-    // resolved different versions at compile-time vs runtime (see #283).
+    // Allowing newer minor versions previously caused NoSuchMethodError due to
+    // compile-time vs runtime version mismatches (see #283).
     implementation 'com.amplitude:analytics-android:1.27.+'
     testImplementation 'org.mockito:mockito-core:4.0.0'
     testImplementation "io.mockk:mockk:1.12.4"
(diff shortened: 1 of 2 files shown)


android/src/main/kotlin/com/amplitude/amplitude_flutter/AmplitudeFlutterPlugin.kt
```diff
@@ -188,45 +189,46 @@ class AmplitudeFlutterPlugin : FlutterPlugin, MethodCallHandler, ActivityAware {
...
-                deepLinks = (map["deepLinks"] as? Boolean) ?: false,
-                // Set false to disable screenViews on Android
-                // screenViews is implemented in Flutter
-                screenViews = false
-            )
+            val sessions = (map["sessions"] as? Boolean) ?: true
+            val appLifecycles = (map["appLifecycles"] as? Boolean) ?: false
+            val deepLinks = (map["deepLinks"] as? Boolean) ?: false
+            // screenViews is always disabled on Android — implemented in Flutter
+            builder.autocapture = buildSet {
```
polbins, line 211: moved usage from `defaultTracking` (deprecated config constructor) to `builder.autocapture`

android/build.gradle
```diff
@@ -46,10 +46,10 @@ dependencies {
     implementation "org.jetbrains.kotlin:kotlin-stdlib-jdk7:$kotlin_version"
     implementation "org.jetbrains.kotlin:kotlin-reflect:1.6.10"
 
-    // Locked to specific version to prevent binary compatibility issues.
+    // Locked to specific minor version to prevent binary compatibility issues.
     // Version range [1.22,2.0) previously caused NoSuchMethodError when Gradle
     // resolved different versions at compile-time vs runtime (see #283).
-    implementation 'com.amplitude:analytics-android:1.25.3'
+    implementation 'com.amplitude:analytics-android:1.27.+'
```
polbins, line 52: I think this is a safe move so that hotfixes are included

android/src/main/kotlin/com/amplitude/amplitude_flutter/AmplitudeFlutterPlugin.kt
chungdaniel: Yep looks good. In the future I'll add `autocapture` as a supported option and mark `defaultTracking` as deprecated on the dart side.

android/build.gradle
chungdaniel: nit: Not sure it's relevant anymore to include the specific version range that the NoSuchMethodError occurred on. Might be better to omit that? Already included context by having the issue #283 mentioned
```
// Locked to specific minor version to prevent binary compatibility issues.
// Allowing newer minor versions previously caused NoSuchMethodError due to
// compile-time vs runtime version mismatches (see #283).
```

chungdaniel (approved):
LGTM!

android/build.gradle
aliaksandr-kazarez: nit: should we cap to `(1.27..2.0]` just in case...?

aliaksandr-kazarez (approved):

PR #305, opened 2026-05-22
feat(config): prefer autocapture over defaultTracking for mobile

chungdaniel:
## Summary

Migrates the Flutter SDK so `Configuration.autocapture` is the canonical autocapture configuration on mobile, matching the native iOS and Android SDKs which deprecated `defaultTracking` some time ago ([Amplitude-Swift Configuration.swift](https://github.com/amplitude/Amplitude-Swift/blob/main/Sources/Amplitude/Configuration.swift), [Amplitude-Kotlin Configuration.kt](https://github.com/amplitude/Amplitude-Kotlin/blob/main/android/src/main/java/com/amplitude/android/Configuration.kt)). The Flutter SDK was the only one still routing mobile config through `defaultTracking`.

- `AutocaptureOptions` now exposes `appLifecycles` and `deepLinks` (mobile). `AutocaptureEnabled` sets them too so the "enable everything" sentinel keeps its semantic.
- `Configuration` constructor's `autocapture` parameter is now nullable. When omitted, an `AutocaptureOptions` is **derived** from the (deprecated) `defaultTracking` field, folding both mobile fields (`sessions`, `appLifecycles`, `deepLinks`) and web fields that exist on both classes (`attribution`, `pageViews`). This is the load-bearing piece: it preserves end-to-end behavior for existing `defaultTracking` callers without requiring any code change on their side.
- Native plugins (Android `AmplitudeFlutterPlugin.kt`, iOS `SwiftAmplitudeFlutterPlugin.swift`) now read only the `autocapture` map and translate it directly to the native `AutocaptureOption` set / `AutocaptureOptions` option set. No merge or fallback logic in native — by the time the map reaches native, Dart has already resolved it.
- `DefaultTrackingOptions` class, `Configuration.defaultTracking` field, and the constructor parameter are all `@Deprecated` with a migration message pointing to `autocapture`.
- iOS `deepLinks` is intentionally not translated — iOS handles deep links outside the native autocapture option set, matching pre-existing behavior on `Amplitude-Swift`.

## Why this approach

A native-side merge (each plugin reads both `autocapture` and `defaultTracking` and resolves) was the first attempt but duplicates the resolution logic across two platforms and leaves `Configuration.autocapture` showing something different from what platforms actually use. Dart-side derivation keeps a single source of truth, makes the native plugins dumb translators, and matches the `didSet` pattern Amplitude-Swift already uses.

It also sets up [planned Flutter frustration tracking](https://amplitude.atlassian.net/wiki/spaces/GOV/pages/3729752561) cleanly: that flag will live on `AutocaptureOptions` but native plugins will explicitly skip it during translation (native frustration tracking only sees `FlutterView` and would emit useless events), and that "skip" lives in one place per platform with no other config to disentangle.

## Backward compatibility

Existing callers using `defaultTracking: DefaultTrackingOptions(...)` see **no behavior change**. The derivation in the constructor folds their values into the autocapture map the native plugin now reads. They'll see one new deprecation warning from `flutter analyze` and a migration message pointing them at `autocapture`.

| Caller | Before | After |
|---|---|---|
| `Configuration(defaultTracking: DTO(appLifecycles: true))` | Mobile reads `defaultTracking` → `{SESSIONS, APP_LIFECYCLES}` | Mobile reads derived `autocapture` → **same** `{SESSIONS, APP_LIFECYCLES}` |
| `Configuration(autocapture: AutocaptureOptions(appLifecycles: true))` | Mobile ignored autocapture; got `{SESSIONS}` from `defaultTracking` default | Mobile reads autocapture → `{SESSIONS, APP_LIFECYCLES}` (now respected) |
| `Configuration(apiKey: 'k')` (neither) | `{SESSIONS}` from `defaultTracking` defaults | Derived autocapture has `sessions: true` → **same** `{SESSIONS}` |
| Both passed | Mobile silently used only `defaultTracking` | `autocapture` wins; `defaultTracking` ignored. New behavior, documented. |

## Test plan

- [x] `flutter test` — all unit tests green (45 tests including 6 new derivation tests)
- [x] `flutter analyze` — clean (only a pre-existing untracked-file issue in scratch)
- [ ] iOS sim manual verification — existing `defaultTracking` config still emits `[Amplitude] Application Started` / Session events
- [ ] iOS sim manual verification — new `autocapture` config emits same events
- [ ] iOS sim manual verification — `AutocaptureEnabled()` enables mobile autocapture
- [ ] iOS sim manual verification — `AutocaptureDisabled()` produces zero `[Amplitude]` events
- [ ] Android emulator manual verification — all four scenarios above
- [ ] Android emulator — deep link via `adb shell am start ...` fires `[Amplitude] Deep Link Opened` when enabled
- [ ] Both passed (`autocapture` + `defaultTracking`) — `autocapture` wins (confirms derivation contract)
- [ ] Dashboard sanity — events from each scenario arrive correctly and look identical between iOS/Android

## Notes for reviewers

- The single `// ignore_for_file: deprecated_member_use_from_same_package` at the top of `lib/configuration.dart`, `test/configuration_test.dart`, and `test/default_tracking_test.dart` is intentional — those files exist to maintain the deprecation bridge and verifying it works. Scattered per-line ignores would be noisier without adding signal.
- `example/pubspec.lock` is removed in a separate `chore:` commit — it was tracked despite `example/pubspec.lock` already being in `.gitignore:21`.
- Kotlin/Swift native tests for autocapture translation weren't added: the existing `AmplitudeFlutterPluginTest.kt` doesn't run via CI (`flutter test` only covers Dart), and the standalone `./gradlew test` build can't resolve `io.flutter.*`. Worth revisiting when Android tests start running through Flutter tooling.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

---

> [!NOTE]
> **Medium Risk**
> Changes which autocapture flags reach native SDKs and introduces explicit `autocapture` vs `defaultTracking` precedence when both are set; behavior for legacy `defaultTracking`-only callers is intended to stay the same via Dart derivation.
> 
> **Overview**
> Makes **`Configuration.autocapture`** the single path for mobile default/autocapture behavior, aligned with native iOS/Android SDKs. **`AutocaptureOptions`** / **`AutocaptureEnabled`** now include **`appLifecycles`** and **`deepLinks`** in their serialized maps.
> 
> **Dart** resolves effective autocapture in **`Configuration._resolveAutocapture`**: explicit **`autocapture`** wins; otherwise values are derived from deprecated **`defaultTracking`** (sessions, mobile flags, and web **`attribution`** / **`pageViews`**). **`defaultTracking`** is **`@Deprecated`** but still emitted in **`toMap()`** for compatibility.
> 
> **Android** and **iOS** plugins stop reading **`defaultTracking`** and map only the **`autocapture`** argument (`false` → disabled; map → native option sets). iOS still does not map **`deepLinks`** from the map (unchanged platform behavior).
> 
> The example app initializes with **`AutocaptureEnabled()`** instead of **`DefaultTrackingOptions.all()`**. Unit tests cover derivation and mobile autocapture serialization.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 87674ffe40e5744df84f9c4c6f4412eb6a15373c. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


chungdaniel:
bugbot run

lib/configuration.dart
```diff
@@ -219,13 +243,37 @@ class Configuration {
...
+  /// the deprecated `DefaultTrackingOptions` when `autocapture` was not passed.
+  ///
+  /// This is the single place where the `defaultTracking → autocapture` bridge
+  /// lives. Native plugins read only the resolved `autocapture` map.
+  static Autocapture _resolveAutocapture(
+    Autocapture? autocapture,
+    DefaultTrackingOptions? defaultTracking,
+  ) {
+    if (autocapture != null) return autocapture;
+    final dt = defaultTracking ?? const DefaultTrackingOptions();
```
aliaksandr-kazarez: nit: dt can't be optional here

example/pubspec.lock
aliaksandr-kazarez, line 1: nit: what the deletion of this file intentional? Wondering what was the issue?

aliaksandr-kazarez (approved):
LGTM;

example/pubspec.lock
chungdaniel: Yes, it was intentional - it was already in the `.gitignore` https://github.com/amplitude/Amplitude-Flutter/blob/main/.gitignore#L21

When changing versions of dependencies it would cause conflicts occasionally, so I figured I might as well just remove it since it's already git ignored as well.

Issue #304, opened 2026-05-14
Request for full Autocapture support across iOS, Android, web


sunweiyang:
## Summary

We have a multi-platform Flutter app (iOS, Android, web) for which we would love to add Amplitude, but Autocapture functionality is currently limited, and only for web. There is a prior ticket for this (#242), but it has been closed. It was mentioned there that this is in your backlog.

## Motivations

Our app is well-developed with a great deal of forms, file upload/download, and interactive tap/click elements. Plus we can really make use of the network and web vitals. Our users span iOS, Android, and web; we need to have full parity of analytics across all platforms. The engineering resources gap to implement Amplitude properly, between Autocapture and manual, is immense.

sunweiyang:
@chungdaniel Let me know if you have questions or need context. We are eager to try this and potentially migrate to Amplitude and away from our other analytics providers if this Autocapture works for our Flutter app!

chungdaniel:
Hey @sunweiyang ! Sorry for the delayed response.

This is still on our backlog, though we likely won’t be able to prioritize the broader full-parity Autocapture work in the near term.

One related update: we recently changed the Flutter SDK so iOS/Android use `autocapture` instead of `defaultTracking` for mobile config. This is mostly a direct port of the existing `defaultTracking` behaviour, since that was already mapping to Autocapture under the hood. So the supported Autocapture capabilities today should be the same as what `defaultTracking` previously provided (appLifecycles, deeplinks [for Android] session tracking), just exposed through the newer/canonical config path. See #305 for details.

What Autocapture areas are you especially interested in for Flutter? While #305 does not fully match functionality with our native iOS, Android, Web SDKs, it can hopefully bridge the gap slightly.

sunweiyang:
@chungdaniel Thanks very much! Form interactions and element interactions are, among other capabilities, will be useful for us.

Would you be open to our team potentially contributing to the Autocapture Flutter parity effort? We can prepare PR(s) for review. Let us know if you're interested!

sunweiyang:
@chungdaniel Our team has taken the liberty of fully implementing remaining Autocapture capabilities. Your team may review and provide feedback for us, if you're interested in adding it to this Flutter SDK!
https://github.com/amplitude/Amplitude-Flutter/pull/309

sunweiyang:
[...]

PR #309, opened 2026-07-09
feat(autocapture): web click/form/download capture, page-URL enrichment, and mobile screen views

sunweiyang:
## Summary

Completes [Amplitude Autocapture](<https://amplitude.com/docs/data/autocapture>) for Flutter, addressing amplitude/Amplitude-Flutter#304. Autocapture was effectively web-only, and even there the Flutter plugin **force-disabled** element clicks, form interactions, and file downloads before handing config to the Browser SDK; mobile had no Flutter route-based screen-view capture.

* **Web (Browser SDK):** element clicks, form interactions, file downloads, and page-URL enrichment are now exposed on `AutocaptureOptions`/`AutocaptureEnabled` and passed through to the Browser SDK instead of being stripped. The newly exposed web options are opt-in by default.
* **Mobile / cross-platform:** screen views are captured via a new `AmplitudeNavigatorObserver`. Installs, upgrades, and sessions already worked through `appLifecycles`/`sessions`.
* **Backward compatibility:** a bare `Configuration(apiKey: ...)` uses the new conservative `AutocaptureOptions` defaults, while explicitly passing the deprecated `defaultTracking` API continues to bridge its legacy values.

### Changes

* Add a sealed `ElementInteractions` type (`ElementInteractionsOptions`/`Enabled`/`Disabled` with `cssSelectorAllowlist`, `actionClickAllowlist`, `dataAttributePrefix`, and `pageUrlAllowlist`), mirroring the existing `Attribution`/`PageViews` pattern.
* Add `formInteractions`, `fileDownloads`, `elementInteractions`, `pageUrlEnrichment`, and `screenViews` to `AutocaptureOptions`/`AutocaptureEnabled`; type `pageViews` as `PageViews`.
* Default `pageViews`, form/file capture, element interactions, page-URL enrichment, and screen views off in `AutocaptureOptions`. `AutocaptureEnabled()` enables all supported options.
* Add a Flutter-aware default `cssSelectorAllowlist`: standard interactive tags plus the common ARIA roles emitted by Flutter’s semantics tree. `ElementInteractionsOptions`, `ElementInteractionsEnabled`, and `AutocaptureEnabled` all use this allowlist.
* Move web config shaping out of `amplitude_web.dart` into a pure-Dart `transformWebConfiguration` helper so it is unit-testable without the `chrome` platform; this removes the forced-off keys and fixes a latent no-op `serverZone.toUpperCase()`.
* Add `AmplitudeNavigatorObserver` (`lib/observers/`): emits `[Amplitude] Screen Viewed` (with `[Amplitude] Screen Name`) on route push/replace/pop, gated on `screenViews`.
* Bridge deprecated `DefaultTrackingOptions` into the resolved `AutocaptureOptions`. Legacy `pageViews` behavior is preserved, while form/file defaults are now false so mobile-only configurations do not opt into newly supported DOM capture.
* Document the Flutter web semantics requirement and its performance caveat in the README, API documentation, and example.
* Bump the example Browser SDK loader to `analytics-browser` 2.44.4 so it exercises every exposed option.

## Why this approach

**Screen views via a Dart** `NavigatorObserver`**, not native** `screenViews`**.** A Flutter app runs inside a single native surface (one `FlutterViewController` / `FlutterActivity`), so the native SDKs' screen-view autocapture only sees that host controller—it cannot observe Flutter route navigation. The observer tracks route changes directly, which is the reliable navigation signal on every platform.

`pageViews` **and screen views remain independent.** On web, a Flutter route change does not necessarily change the browser URL, so the route-based observer captures navigations the Browser SDK's URL-based `pageViews` can miss. Both default off in the new API to avoid unexpectedly double-counting navigation. Applications can enable either or both depending on their routing and analytics requirements.

Explicit legacy `defaultTracking` configurations continue to preserve their existing `pageViews` value. This is intentional: page views were previously supported and default-on through that API.

**Web config shaping is pure Dart.** `amplitude_web.dart` imports `dart:js_interop` and is excluded from analysis / VM tests, so the logic lives in `transformWebConfiguration` (fully unit-tested) with only the trivial `.jsify()` left in the plugin.

## Backward compatibility

**Behavior change (web):** a bare `Configuration(apiKey: ...)` and `AutocaptureOptions()` now disable URL-based page-view capture by default. Page views previously defaulted on, so applications relying on `[Amplitude] Page Viewed` events should explicitly enable `PageViewsOptions()`.

The newly exposed form, file, element, and page-URL features default off. Upgrading therefore does not begin capturing these new event types unless the application opts in or uses `AutocaptureEnabled()`.

| Caller | Before | After |
| -- | -- | -- |
| Bare `Configuration(apiKey: ...)` | Page views on; form/file/click force-disabled | Page views and newly exposed web options off |
| Web `AutocaptureOptions()` | Page views on; form/file/click force-disabled | Page views and newly exposed web options off |
| Web `AutocaptureEnabled()` | Form/file/click force-disabled | All supported options enabled, with Flutter-aware click selectors |
| Explicit legacy `defaultTracking` | Page views follow legacy value; form/file stripped | Page views still follow legacy value; explicit form/file opt-ins are honored |
| Mobile with `screenViews: true` and observer attached | Not captured | Captured via `AmplitudeNavigatorObserver` |

## Web integration notes (README/example)

* **Per-option Browser SDK minimums.** Full autocapture needs `elementInteractions` >= **2.10.0**, while `pageUrlEnrichment` is silently ignored below **2.29.0**. The README documents the per-option minimums and the example uses **2.44.4**.
* **Flutter-web (CanvasKit) DOM-capture limitation.** The Browser SDK's DOM-based capture (`elementInteractions`, `formInteractions`, `fileDownloads`) can only see real DOM nodes. In practice, capture works through Flutter’s accessibility semantics tree when semantics are enabled and not otherwise.
* **Semantics tradeoffs.** Enabling Flutter semantics globally may affect runtime performance and behavior. DOM-based capture is therefore opt-in and the requirement is documented next to the relevant options.

Web consumers must load the Browser SDK snippet in their own `web/index.html`; this is documented in the README, and web tests run via `flutter test --platform chrome`.

## Test plan

- [X] `flutter test` — Dart unit and widget tests green
- [X] `flutter test --platform chrome` — web `jsify()` glue test green
- [X] `flutter analyze` / `dart analyze lib test` — clean
- [X] Configuration serialization and deprecated DTO bridge coverage
- [X] Flutter-aware selector and enabled-sentinel coverage
- [X] Navigator observer push/replace/pop, dialog dedupe, initialization, and error-isolation coverage
- [X] Web live test against `analytics-browser` 2.44.4
- [X] Android emulator live test
- [X] iOS simulator live test
- [ ] Native Kotlin/Swift CI — see notes

## Notes for reviewers

* **Scope:** sessions, page views, clicks, form interactions, file downloads, marketing attribution, and page-URL enrichment on web; installs, upgrades, sessions, and Flutter route-based screen views on mobile. `networkTracking`, `frustrationInteractions`, and `webVitals` are intentionally out of scope.
* **No native unit tests added.** The new web keys are Dart/web concerns and the native plugins deliberately ignore them. Screen views come from the Dart observer and are covered by unit/widget and live platform tests.
* **Modeling:** `formInteractions`/`fileDownloads` are simple `bool`s; `elementInteractions.pageUrlAllowlist` is String-only because JS `RegExp` and `shouldTrackEventResolver` cannot cross the platform channel.
* **Release note:** the default `pageViews` change should be called out because bare/new configurations now require an explicit opt-in to continue emitting `[Amplitude] Page Viewed`.


lib/autocapture/autocapture.dart
```diff
@@ -65,11 +66,12 @@ class AutocaptureOptions extends Autocapture {
...
-  /// Configuration for autocapturing page view events using `PageViewsOptions`. See [docs](https://amplitude.com/docs/sdks/analytics/browser/browser-sdk-2#track-page-views)
-  /// for more information. Set to `false` to disable tracking page views.
+  /// Configures autocapturing page view events. See [docs](https://amplitude.com/docs/sdks/analytics/browser/browser-sdk-2#track-page-views)
+  /// for more information.
   ///
-  /// Can be either `PageViewsOptions` or `false`.
-  final dynamic pageViews;
+  /// Use [PageViewsOptions] for granular control, or [PageViewsEnabled]/
+  /// [PageViewsDisabled] to toggle it.
+  final PageViews pageViews;
```
chungdaniel, line 82: nit: Given that `PageViews` has overlap with `ScreenViews` and ScreenViews seem more consistent with Flutter and the other platforms, I'd rather disable PageViews by default.

lib/autocapture/autocapture.dart
```diff
@@ -82,12 +84,60 @@ class AutocaptureOptions extends Autocapture {
...
+  /// a Flutter app runs inside a single native surface (one `FlutterViewController`
+  /// on iOS, one `FlutterActivity` on Android).
+  final bool screenViews;
+
+  /// Web specific
+  ///
+  /// Whether to capture form interaction events (`[Amplitude] Form Started`,
+  /// `[Amplitude] Form Submitted`).
+  ///
+  /// Exposed as a simple on/off toggle; unlike [elementInteractions], the
```
chungdaniel, line 111: For `formInteractions`, `fileDownloads`, and `elementInteractions`, it'd be helpful if the requirement that semantics/semantic tagging be enabled was documented more explicitly and in more places, as it's an easy thing to miss right now (i.e. more than just in the README).

lib/autocapture/autocapture.dart
```diff
@@ -82,12 +84,60 @@ class AutocaptureOptions extends Autocapture {
...
+  final bool pageUrlEnrichment;
+
   const AutocaptureOptions({
     this.attribution = const AttributionOptions(),
     this.sessions = true,
     this.pageViews = const PageViewsOptions(),
     this.appLifecycles = false,
     this.deepLinks = false,
+    this.screenViews = false,
+    this.formInteractions = true,
```
chungdaniel: Would prefer the new web autocapture features to not be enabled by default due to the semantics requirement

lib/autocapture/element_interactions.dart
```diff
@@ -0,0 +1,132 @@
...
+///         ),
+///     )
+/// );
+/// ```
+class ElementInteractionsOptions extends ElementInteractions {
+  /// Web specific
+  ///
+  /// List of CSS selectors that gate which elements are tracked. When set, only
+  /// elements matching a selector generate `[Amplitude] Element Clicked` events.
+  final List<String>? cssSelectorAllowlist;
```
chungdaniel, line 92: In my testing, I found we had to explicitly add entries to cssSelectorAllowList to get things working, e.g. '[role="button"]'. This seems like it might be a common pattern for Flutter apps generally. Would it be possible to have cssSelectorAllowList include common Flutter semantic selectors by default? Did you have to do anything specific to get this feature working?

lib/autocapture/autocapture.dart
chungdaniel: Additionally, since semantics is required, I looked into it a bit. Seems like it has some known sideeffects and performance degradations. Think we should have some sort of caveat, e.g.

> Note: `elementInteractions` requires Flutter semantics to be enabled globally. This may affect app performance and is not recommended unless your app already relies on semantics.

chungdaniel (changes requested):
Hi @sunweiyang, sorry for the delayed review, this one took longer to get to than it should have. Appreciate your patience, and thanks for the contribution!

Overall the PR looks great! Just a few small comments here and there. Curious if you needed to do anything else in your implementation that you think may carry to all Flutter apps in general (e.g. like the `cssSelectorAllowlist` comment) to get these new features to work?

lib/autocapture/autocapture.dart
sunweiyang: Agreed, and done. `pageViews` now defaults to `PageViewsDisabled()`. Navigation is captured cross-platform by `screenViews` + `AmplitudeNavigatorObserver`, so the web URL-based page views are opt-in and a navigation doesn't get double-counted.

One thing to flag: the SDK previously defaulted `pageViews` on, so this is a behavior change for web-only callers, who'll stop getting `[Amplitude] Page Viewed` unless they set `pageViews: PageViewsOptions()`. I've noted it under an "Upgrading" section in the README. Happy to leave it on if you'd rather not change that default.

lib/autocapture/autocapture.dart
sunweiyang: Done, and thanks for chasing down the performance angle. The requirement is now documented in a few places instead of just the README: the `elementInteractions` dartdoc (with `formInteractions`/`fileDownloads` pointing to it), `ElementInteractionsOptions`, and a dedicated README "DOM-based capture on Flutter web" section. I used your caveat almost as written: each spot notes that semantics is app-wide, has a runtime cost and known side effects, and isn't recommended unless the app already relies on it. The example's `main()` also shows the `ensureSemantics()` call, guarded to web.

lib/autocapture/autocapture.dart
sunweiyang: Agreed. `formInteractions`, `fileDownloads`, `elementInteractions`, `pageUrlEnrichment`, and `pageViews` all default off now. That also drops the "form/file default-on" behavior change the original PR had flagged for a release call, since upgrading web apps won't capture anything new unless they opt in.

One subtlety I had to fix for this to actually hold: a bare `Configuration(apiKey: ...)` derives `autocapture` from `DefaultTrackingOptions`, so I made that path return a default `AutocaptureOptions` when neither `autocapture` nor `defaultTracking` is passed. An explicit `defaultTracking:` still bridges through for backward compatibility.

lib/autocapture/element_interactions.dart
sunweiyang: Yeah, I ran into the same thing, and baking it in makes sense. `ElementInteractionsOptions` now defaults `cssSelectorAllowlist` to `ElementInteractionsOptions.defaultCssSelectorAllowlist`: the standard interactive tags plus the semantic roles Flutter emits (`[role="button"]`, `[role="link"]`, `[role="checkbox"]`, `[role="radio"]`, `[role="switch"]`, `[role="tab"]`, `[role="menuitem"]`, `[role="textbox"]`). So `ElementInteractionsOptions()` captures Flutter widgets without hand-tuning; pass your own list to replace it, or `cssSelectorAllowlist: null` to fall back to the Browser SDK's default.

To your question about what else was needed: the other half is enabling the semantics tree globally, since otherwise there are no role nodes to match at all. That's in the README and example now.

Two things I'd like your read on:
- The role list is just the common interactive roles. If your testing turned up others worth including by default, or if `[role="textbox"]` is redundant with `input` once semantics is on, I'll adjust.
- `AutocaptureEnabled()` still serializes `elementInteractions: true` (the raw Browser SDK default, no Flutter roles), so it won't capture Flutter clicks the way `ElementInteractionsOptions()` does. I left it as a literal passthrough to keep the sentinel simple, but I can have it emit the Flutter-aware allowlist if you'd rather `AutocaptureEnabled()` just work on web.

sunweiyang:
Good question. Two things carry to basically every Flutter web app, and I've handled both in the SDK/docs so integrators don't have to rediscover them:

1. The accessibility semantics tree has to be enabled globally (`SemanticsBinding.instance.ensureSemantics()`). With CanvasKit there's no real DOM for the Browser SDK to see until it's on. That's now on the option dartdocs and in the README with a performance caveat, plus shown in the example's `main()`.
2. Flutter emits ARIA roles instead of native tags, so the Browser SDK's default `cssSelectorAllowlist` never matches Flutter's `<flt-semantics role="...">` nodes. `ElementInteractionsOptions` now defaults its allowlist to a Flutter-aware set that includes those roles.

On mobile the only piece needed was the `AmplitudeNavigatorObserver`, which is already in this PR. I've also made all web autocapture options opt-in (off by default) per the comments below.

lib/autocapture/element_interactions.dart
chungdaniel: 1. The role list you specified looks great, don't recall having extras that you didn't already have.
2. I think it would definitely make sense to have it emit the Flutter-aware allowlist. For most (?) users that don't have semantics enabled, it wouldn't result in new element interaction events being captured which should be ok. Would still be in the spirit of `AutocaptureEnabled`.

lib/configuration.dart
```diff
@@ -275,6 +280,8 @@ class Configuration {
       pageViews: defaultTracking.pageViews
           ? const PageViewsOptions()
           : const PageViewsDisabled(),
+      formInteractions: defaultTracking.formInteractions,
+      fileDownloads: defaultTracking.fileDownloads,
```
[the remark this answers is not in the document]
chungdaniel: This is a valid case (though in most cases, as stated above, wouldn't impact most users due to requirement of semantics). Was something on my end prior to your autocapture PR so I'm happy to address it myself after any extra changes (just need to change defaultTracking's default values for formInteractions and fileDownloads to false rather than true..)

chungdaniel (changes requested):
One small change request on your end regarding emitting Flutter-aware allow list for cssSelectorAllowList and we should be good to merge!

lib/autocapture/element_interactions.dart
sunweiyang: Thanks, and good call. This is already in as of 169a826: both `ElementInteractionsEnabled` and `AutocaptureEnabled` now emit the Flutter-aware `defaultCssSelectorAllowlist` instead of a bare `true`, so "enable everything" captures Flutter web clicks. See `ElementInteractions.toMapOrBool` in element_interactions.dart and `AutocaptureEnabled.toMap` in autocapture.dart. Agree it stays in the spirit of `AutocaptureEnabled`, and as you note it's a no-op for apps without semantics enabled.

lib/configuration.dart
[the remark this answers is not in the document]
sunweiyang: I went ahead and made this change in 5976cd4 to save you the round-trip: `DefaultTrackingOptions.formInteractions`/`fileDownloads` now default to `false` (the `.all()` factory and explicit opt-ins are unchanged), so the bridge no longer enables web form/file capture for callers that only set mobile fields. Added a regression test. Happy to drop it if you'd rather own that change on your end.

lib/default_tracking.dart
```diff
@@ -37,8 +37,11 @@ class DefaultTrackingOptions {
     this.deepLinks = false,
     this.attribution = true,
     this.pageViews = true,
-    this.formInteractions = true,
-    this.fileDownloads = true,
+    // Web DOM-based capture is opt-in (see AutocaptureOptions). Defaulting these
```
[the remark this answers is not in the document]
chungdaniel: This is ok, default tracking is already deprecated, so this is just to maintain backwards compatibility.

chungdaniel (approved):

chungdaniel:
Updated the PR description to reflect the final opt-in defaults and legacy defaultTracking behavior. Thanks, everything looks good now!

